PR #4, opened 2016-08-16
Fix order of application for migrations

amacneil:
Go iterates through map keys in random order, causing migrations to also be applied in random order. I also added an explicit sort in case `ioutil.ReadDir()` gives us files in non-alphabetic order.

@defunctzombie @dthakur


defunctzombie:
amazing. LGTM

PR #8, opened 2017-03-01
Add musl target for Alpine support

amacneil:
Building with `CGO_ENABLED=0` results in a binary which can run on Alpine linux (with musl libc), but it also disables SQLite support.

Adding an additional target allows people to decide this tradeoff for themselves.


amacneil:
@defunctzombie

defunctzombie:
:partyparrot:

defunctzombie:
:+1:

defunctzombie:
might be nice to add something to the readme about it

amacneil:
Updated readme.

PR #9, opened 2017-04-17
Refactor dbmate package to be importable as a package

flowhamster:
Refactored dbmate to allow gocode to include dbmate as a vendor dependency. I wanted to make this change so that my golang tests could setup migrate and teardown database dependencies.

I realize that to do this pull request we will need to fix the package paths in the import headers on several files. Should be an easy fix. I did this as a fork in case so that I could reference this library while including any additional changes that you want.


amacneil:
👍 like the direction. I added you as collaborator so you can push it to a branch in this repo and test fixing up the import paths.

Is there any good argument for/against having all the separate packages (e.g. separate package for each driver, commands, utils, version etc)? If you look in the git history I actually started out having separate packages for each driver, but ended up moving to a flat structure because it seemed overkill.

soheil:
Would be interested to learn the pros/cons of having separate packages too, specially for smaller projects.

flowhamster:
There are a few different reasons and thoughts to separate things out into packages. From an IDE perspective smaller packages mean that you are exposing less in your API, which makes it a little easier to work develop with. It also reduces the vendor/build bloat of other projects that might include your project since a go build wouldnt necessarily have to include all the vendor dependencies that `dbmate/cmd/dbmate` would.

amacneil:
That makes sense, but in something this small, is there value in having things broken out to the point of having a separate `version` package? It seems overkill. Likewise, the drivers are pretty close to useless on their own, so they may as well just be part of the main package too.

flowhamster:
Created a new PR with the suggestion. In something this small it really doesnt matter. Let me know if the changes in #10 better reflect what you are thinking.

PR #15, opened 2017-05-22
Add support for multiple projects

kfr2:
This change supports storing various sets of migrations under
different, independently-deployable projects. For example, you may use
this tool with the projects app1 and app2. These projects are
responsible for maintaining their own sets of database tables and
should not worry about any other migrations that may have occured.
Thus, the -p (--project) flag can be passed to the dbmate CLI to
restrict its operations to any migrations that may have occurred (or
will need to occur) for the particular project.

For example, usage may look like:
```
/home $ ./dbmate -d app1 -p app1 migrate
Applying: 1495196229_create_users_table.sql
/home $ ./dbmate -d app2 -p app2 migrate
Applying: 1495197370_create_metrics_table.sql
/home $ ./dbmate -d app1 -p app1 down
Rolling back: 1495196229_create_users_table.sql
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ Rolling back: 20151127184807_create_users_table.sql
 The following command line options are available with all commands. You must use command line arguments in the order `dbmate [global options] command [command options]`.
 
 * `--migrations-dir, -d "./db/migrations"` - where to keep the migration files.
+* `--project, -p "project-name"` - a name under which to associate the set of migrations. defaults to 'default'
 * `--env, -e "DATABASE_URL"` - specify an environment variable to read the database connection URL from.
 
 For example, before running your test suite, you may wish to drop and recreate the test database. One easy way to do this is to store your test database connection URL in the `TEST_DATABASE_URL` environment variable:
--- a/cmd/dbmate/main.go
+++ b/cmd/dbmate/main.go
@@ -40,6 +40,11 @@ func NewApp() *cli.App {
 			Value: "DATABASE_URL",
 			Usage: "specify an environment variable containing the database URL",
 		},
+		cli.StringFlag{
+			Name: "project, p",
+			Value: "default",
+			Usage: "specify a name to associate with the migration set",
+		},
 	}
 
 	app.Commands = []cli.Command{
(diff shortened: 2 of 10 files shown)


amacneil:
Hey, thanks for the PR!

Can you explain the use case a little more? I don't think it's a problem I've run into in the past. I guess I have two main questions:

* In this situation are you expecting these multiple "projects" to share the same database? In that case are they not really one project?
* Would it make more sense to just store the migration files for each project under a separate folder? For example:

```
/myapp
    /project1
      /db
        /migrations
    /project2
      /db
        /migrations
```

kfr2:
I'd like to use dbmate to manage migrations for a number of microservices, each of which is responsible for maintaining its own migration files. Each of these projects is also contained within its own repository. For example, the repositories might be structured like:

```
/app1
    api.py
    migrations/
         1495196229_create_users_table.sql
         ...

/app2
    api.py
    models.py
    migrations/
        1495197370_create_food_table.sql
         ...

/app3
    api.py
    migrations/
        1395196229_create_books_table.sql
        ...
```

It seems that dbmate currently requires all the migration files it is managing to be located in the same directory -- especially when attempting to reverse a migration. For example, if I run the migrations for the projects listed above using dbmate's latest master branch, I'm unable to revert app3's migration because it needs the knowledge of the migrations that appear "after" it based on the file timestamp. 

```
/home # ./dbmate -d app1 up
Applying: 1495196229_create_users_table.sql
/home # ./dbmate -d app2 up
Applying: 1495197370_create_food_table.sql
/home # ./dbmate -d app3 up
Applying: 1395196229_create_books_table.sql
/home # ./dbmate -d app3 down
can't find migration file: 1495197370*.sql
```

This change namespaces migrations to a particular project name such that dbmate only cares about the migration files local to a particular repository.

```
/home # ./dbmate -p app1 -d app1 up
Applying: 1495196229_create_users_table.sql
/home # ./dbmate -p app2 -d app2 up
Applying: 1495197370_create_food_table.sql
/home # ./dbmate -p app3 -d app3 up
Applying: 1395196229_create_books_table.sql
/home # ./dbmate -p app3 -d app3 down
Rolling back: 1395196229_create_books_table.sql
```

If the *-p* flag is not passed, dbmate should continue to work as it currently does by associating everything with the 'default' project.

amacneil:
Are each of your services talking to the same database, or different databases?

If they are different databases, then I think this would be more easily solved by simply having a different `.env` / connection string for each service.

If they are all connecting to the same database, then I wonder what the value of keeping the migration files separated under each service is? If they are all sharing the same database, then you can not fully reason about the migrations being independent of each other, it might be simpler to keep them all in a single "database" project that is managed independently of the specific services.

PR #16, opened 2017-05-25
Add darwin build target

kfr2:
Running `make build` will produce 4 binaries in the *dist* directory. This change adds darwin (OS X) as a build target.


amacneil:
Did you get this working? I actually originally had some macos build scripts, but gave up in favor of the homebrew buidls.

PR #32, opened 2018-03-10
feature: Proposal of status command

hmleal:
Please this is just a proposal about the ```status``` command in dbmate. Do you could help with the tests? I'll try as well

And thank you for this amazing project

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ dbmate migrate   # run any pending migrations
 dbmate rollback  # roll back the most recent migration
 dbmate down      # alias for rollback
 dbmate dump      # write the database schema.sql file
+dbmate status    # show the status of all migrations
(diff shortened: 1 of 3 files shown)


MarkMurphy:
Would be nice if you could pass a flag that would tell this to simply output the number of pending migrations or exit with a non-zero status if there are any. This would be really useful for external applications or scripts.

fernandocarletti:
@amacneil Is there a reason for not having this merged into dbmate? It is a helpful command to check if there's nothing pending :)

amacneil:
@fernandocarletti definitely agree I would like to add this command to dbmate. This PR is just missing some tests and cleanup. I haven't had time to do it myself but if someone wants to contribute I would gladly accept it.

PR #45, opened 2018-06-06
Feature: Support testing if migrations will succeed 

vrajanap:
This will allow user to test if the set of migrations will apply correctly.

This will be allow clients to use dbmate and test if migrations will succeed across multiple databases before going ahead and apply on them.
#### TODO

- [x]  Update README
- [x]  Got unit tests.
- [x]   Update dbmate documentation.

### DIFF
--- a/main.go
+++ b/main.go
@@ -74,7 +74,10 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				return db.CreateAndMigrate(c.Bool("with-rollback"))
+				if c.Bool("with-rollback") {
+					return db.CreateAndMigrateWithRollback()
+				}
+				return db.CreateAndMigrate()
 			}),
 		},
 		{
--- a/main.go
+++ b/main.go
@@ -64,8 +64,17 @@ func NewApp() *cli.App {
 		{
 			Name:  "up",
 			Usage: "Create database (if necessary) and migrate to the latest version",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name: "with-rollback",
+					Usage: "Checks if this command can succeed by running migrations " +
+						"within a transaction that is ultimately rolled back\n" +
+						"\tWARNING:- MySQL doesn't support transactional DDL and " +
+						"therefore this flag results in an error for MySQL",
+				},
+			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				return db.CreateAndMigrate()
+				return db.CreateAndMigrate(c.Bool("with-rollback"))
 			}),
 		},
 		{
@@ -85,8 +94,17 @@ func NewApp() *cli.App {
 		{
 			Name:  "migrate",
 			Usage: "Migrate to the latest version",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name: "with-rollback",
+					Usage: "Checks if this command can succeed by running migrations " +
+						"within a transaction that is ultimately rolled back\n" +
+						"\tWARNING:- MySQL doesn't support transactional DDL and " +
+						"therefore this flag results in an error for MySQL",
+				},
+			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				return db.Migrate()
+				return db.Migrate(c.Bool("with-rollback"))
 			}),
 		},
 		{
--- a/main.go
+++ b/main.go
@@ -51,7 +51,7 @@ func NewApp() *cli.App {
 		},
 	}
 
-	withRollbackFlag := "with-rollback"
+	const withRollbackFlag = "with-rollback"
 	app.Commands = []cli.Command{
 		{
 			Name:    "new",
--- a/main.go
+++ b/main.go
@@ -110,9 +110,8 @@ func NewApp() *cli.App {
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
 				if c.Bool(withRollbackFlag) {
 					return db.MigrateWithRollback()
-				} else {
-					return db.Migrate()
 				}
+				return db.Migrate()
 			}),
 		},
 		{
--- a/main.go
+++ b/main.go
@@ -51,6 +51,7 @@ func NewApp() *cli.App {
 		},
 	}
 
+	withRollbackFlag := "with-rollback"
 	app.Commands = []cli.Command{
 		{
 			Name:    "new",
@@ -66,7 +67,7 @@ func NewApp() *cli.App {
 			Usage: "Create database (if necessary) and migrate to the latest version",
 			Flags: []cli.Flag{
 				cli.BoolFlag{
-					Name: "with-rollback",
+					Name: withRollbackFlag,
 					Usage: "Checks if this command can succeed by running migrations " +
 						"within a transaction that is ultimately rolled back\n" +
 						"\tWARNING:- MySQL doesn't support transactional DDL and " +
@@ -74,7 +75,7 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				if c.Bool("with-rollback") {
+				if c.Bool(withRollbackFlag) {
 					return db.CreateAndMigrateWithRollback()
 				}
 				return db.CreateAndMigrate()
@@ -99,7 +100,7 @@ func NewApp() *cli.App {
 			Usage: "Migrate to the latest version",
 			Flags: []cli.Flag{
 				cli.BoolFlag{
-					Name: "with-rollback",
+					Name: withRollbackFlag,
 					Usage: "Checks if this command can succeed by running migrations " +
 						"within a transaction that is ultimately rolled back\n" +
 						"\tWARNING:- MySQL doesn't support transactional DDL and " +
@@ -107,7 +108,11 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				return db.Migrate(c.Bool("with-rollback"))
+				if c.Bool(withRollbackFlag) {
+					return db.MigrateWithRollback()
+				} else {
+					return db.Migrate()
+				}
 			}),
 		},
 		{
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -303,8 +303,8 @@ func (db *DB) migrateAndRollback() error {
 	defer mustClose(sqlDB)
 
 	if !drv.SupportsTransactionalDDL() {
-		return fmt.Errorf("%s doesn't support Transactional DDL. "+
-			"Therefore, with-rollback flag is not supported", db.DatabaseURL.Scheme)
+		return fmt.Errorf("%s doesn't support Transactional DDL. Therefore, with-rollback flag is not supported",
+			db.DatabaseURL.Scheme)
 	}
 
 	applied, err := drv.SelectMigrations(sqlDB, -1)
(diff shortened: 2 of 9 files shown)


vrajanap:
@amacneil Have you had a chance to take a look? Or have you picked anyone as maintainer to review this?

pkg/dbmate/db_test.go
```diff
@@ -86,7 +86,7 @@ func TestDumpSchema(t *testing.T) {
 	require.NoError(t, err)
 
 	// create and migrate
-	err = db.CreateAndMigrate()
+	err = db.CreateAndMigrate(false)
```
amacneil: In general I'm not a fan of boolean args like this (see https://www.martinfowler.com/bliki/FlagArgument.html), since it's not obvious to the caller what the arg represents. Could we split this into separate methods?

pkg/dbmate/db.go
```diff
@@ -85,28 +85,44 @@ func (db *DB) Wait() error {
 	return fmt.Errorf("unable to connect to database: %s", err)
 }
 
-// CreateAndMigrate creates the database (if necessary) and runs migrations
-func (db *DB) CreateAndMigrate() error {
+// CreateAndMigrate creates the database (if necessary) and runs migrations.
+// If withRollback is true, this doesn't create or migrate but
+// returns if creation of database and applying of migrations would succeed.
+func (db *DB) CreateAndMigrate(withRollback bool) error {
```
bradclawsie: Just create a new function for this. Someone mis-placing the arg here could drop their DB. Probably not what they want. The arg is misleading because dropping a DB cannot be rolled back on any RDBMS I know of.

The goal seems to be to experiment with running a migration. A safer way to do that is to dump the CREATE statement for the current schema (typically requires querying metadata tables), and then creating a temporary table with a synthesized name and no rows...run the migration on *that* if you want to drop it. You probably never want to hide a DROP inside a function if a user actually has data in a table etc.

pkg/dbmate/db.go
```diff
@@ -202,6 +218,23 @@ func doTransaction(db *sql.DB, txFunc func(Transaction) error) error {
 	return tx.Commit()
 }
 
+func transactAndRollback(db *sql.DB, txFunc func(Transaction) error) error {
+	tx, err := db.Begin()
```
bradclawsie: db could be nil, this would panic

pkg/dbmate/db.go
```diff
@@ -202,6 +218,23 @@ func doTransaction(db *sql.DB, txFunc func(Transaction) error) error {
...
+	if err != nil {
+		return err
+	}
+
+	if err := txFunc(tx); err != nil {
```
bradclawsie: YMMV, but it seems to be useful to preserve the transaction error message even if the rollback errors out, both are non-nil...when I have coded this same section in the past, I have done something like

```
return fmt.Errorf("Rollback err %s on transaction err %s",err1.Error(),err.Error())
```

pkg/dbmate/db.go
```diff
@@ -221,8 +254,67 @@ func (db *DB) openDatabaseForMigration() (Driver, *sql.DB, error) {
 	return drv, sqlDB, nil
 }
 
-// Migrate migrates database to the latest version
-func (db *DB) Migrate() error {
+// Migrate migrates database to the latest version. If withRollback is true,
+// this doesn't migrate but returns if such a migration may fail.
+func (db *DB) Migrate(withRollback bool) error {
+	if withRollback {
```
bradclawsie: Just make a new toplevel func
 
```MigrateAndDrop```

File: pkg/dbmate/db.go [hunk 6]
```diff
@@ -221,8 +254,67 @@ func (db *DB) openDatabaseForMigration() (Driver, *sql.DB, error) {
...
+	err = transactAndRollback(sqlDB, func(tx Transaction) error {
+		for _, filename := range files {
+			ver := migrationVersion(filename)
+			if ok := applied[ver]; ok {
+				// migration already applied
+				continue
+			}
+
+			fmt.Printf("Attempting to apply : %s\n", filename)
+			migration, err := parseMigration(filepath.Join(db.MigrationsDir, filename))
```
bradclawsie: +1 for not just hardcoding '/'

pkg/dbmate/db.go
```diff
@@ -85,28 +85,44 @@ func (db *DB) Wait() error {
...
+// If withRollback is true, this doesn't create or migrate but
+// returns if creation of database and applying of migrations would succeed.
+func (db *DB) CreateAndMigrate(withRollback bool) error {
 	drv, err := db.GetDriver()
 	if err != nil {
 		return err
 	}
 
+	if !drv.SupportsTransactionalDDL() && withRollback {
+		return fmt.Errorf("%s doesn't support Transactional DDL. "+
```
bradclawsie: No need for string concat here, just make it one long string

pkg/dbmate/db.go
```diff
@@ -221,8 +254,67 @@ func (db *DB) openDatabaseForMigration() (Driver, *sql.DB, error) {
...
+
+func (db *DB) migrateAndRollback() error {
+	re := regexp.MustCompile(`^\d.*\.sql$`)
+	files, err := findMigrationFiles(db.MigrationsDir, re)
+	if err != nil {
+		return err
+	}
+
+	if len(files) == 0 {
+		return fmt.Errorf("no migration files found")
```
bradclawsie: can just use 
```
errors.New()
```

pkg/dbmate/db.go
```diff
@@ -202,6 +218,23 @@ func doTransaction(db *sql.DB, txFunc func(Transaction) error) error {
...
 }
 
+func transactAndRollback(db *sql.DB, txFunc func(Transaction) error) error {
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+
+	if err := txFunc(tx); err != nil {
+		if err1 := tx.Rollback(); err1 != nil {
```
bradclawsie: There's nothing in the function signature for ```txFunc``` that tells me the relationship between getting an ```error``` back and the transaction state (did it ```Commit```? Who knows)...Seems like this is just easier to inline where you need it to avoid this

pkg/dbmate/db_test.go
vrajanap: Good point. Will do.

main.go
```diff
@@ -64,8 +64,17 @@ func NewApp() *cli.App {
 		{
 			Name:  "up",
 			Usage: "Create database (if necessary) and migrate to the latest version",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name: "with-rollback",
```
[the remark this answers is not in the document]
vrajanap: Good point, will do.

pkg/dbmate/db.go
vrajanap: Good points. I will split this into a separate function. 

Btw, I want to clarify that this function doesn't support `MigrateAndDrop`. It only drops if it creates the database here _and_ `withRollback` was true.

vrajanap: If you feel that the design of even supporting withRollback for Creation part of Create and migrate is dangerous, I agree and I'm willing to remove that functionality. For my scenario, I only care about "withRollback" for migrations and not at all about creation.

pkg/dbmate/db.go
vrajanap: Love this suggestion!

vrajanap: This is a good idea. Will do.

pkg/dbmate/db.go
vrajanap: golint doesn't seem to like lines longer than 100 columns.

pkg/dbmate/db.go
```diff
@@ -202,6 +218,23 @@ func doTransaction(db *sql.DB, txFunc func(Transaction) error) error {
 	return tx.Commit()
 }
 
+func transactAndRollback(db *sql.DB, txFunc func(Transaction) error) error {
```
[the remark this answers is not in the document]
vrajanap: I don't quite follow your question. Transaction is a function that is being executed, whereas sql.Tx is a db transaction object that can be used to invoke the transaction.

pkg/dbmate/db.go
vrajanap: Good point, but this function is used only from one location where param is guaranteed to be not nil. I am not familiar with golang patterns here, but if we start checking all param in each level, wouldn't that lead to lot of duplicate checks? :thinking:

pkg/dbmate/db.go
vrajanap: This was old behavior as in https://github.com/amacneil/dbmate/blob/master/pkg/dbmate/db.go#L233

Won't changing this impact other scripts if they were greping for this?

vrajanap:
@amacneil @bradclawsie PTAL. Thanks for the great feedback. Please keep it coming. :)

pkg/dbmate/db_test.go
vrajanap: done.

main.go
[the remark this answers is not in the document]
vrajanap: done

pkg/dbmate/db.go
vrajanap: done.

pkg/dbmate/db.go
vrajanap: done

pkg/dbmate/db.go
vrajanap: done

pkg/dbmate/db.go
vrajanap: thanks! :)

pkg/dbmate/db.go
vrajanap: Lemme know if there is someway around golint rules

pkg/dbmate/db.go
bradclawsie: yeah don't worry about it this is very minor

pkg/dbmate/db.go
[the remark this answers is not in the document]
bradclawsie: no worries, I think I was just trying to figure out why `Transaction` existed

pkg/dbmate/db.go
vrajanap: :man_shrugging:

pkg/dbmate/db.go
vrajanap: Yea, your other suggestions for `transactAndRollback` also makes sense for `doTransaction`. I will make this change.

vrajanap:
@bradclawsie Noted all your comments. I just made one minor change following your recommendation for `doTransaction`. Let me know if the changes look good. Eager to merge this PR! :)

pkg/dbmate/db.go
```diff
@@ -106,7 +107,43 @@ func (db *DB) CreateAndMigrate() error {
...
+		if err := drv.CreateDatabase(db.DatabaseURL); err != nil {
+			return err
+		}
+		createdDb = true
+	}
+
+	// migrate
+	err = db.MigrateWithRollback()
+	if err == nil && createdDb {
+		return drv.DropDatabase(db.DatabaseURL)
```
amacneil: I would leave this whole step out (dropping the db), I don't think it's necessary in the context of this flag.

(I'm happy if we just create the db, test migration, then roll back migration, and leave the empty db existing)

vrajanap: :+1: WIll fix remove the dropping piece of it.

amacneil:
Closing stale PR since we are not currently planning to integrate this. I'm not convinced that this is a safe operation to run against a production database (even within a transaction), and given that we now support disabling transactions for postgres (and transactions are not supported for mysql) this feature would have unexpected behavior in a number of situations, potentially leading to data loss or downtime if someone misunderstood how it works.

PR #49, opened 2018-07-22
Add path to schema_migrations insert statement

pkqk:
Currently importing `schema.sql` fails when inserting the current migrations into the `schema_migrations` table.

At the top of the output of pg_dump it adds the line:

```sql
SELECT pg_catalog.set_config('search_path', '', false);
```

Which wipes out the schema `search_path` so every table reference needs to be absolute.

### DIFF
--- a/pkg/dbmate/postgres.go
+++ b/pkg/dbmate/postgres.go
@@ -79,7 +79,7 @@ func postgresSchemaMigrationsDump(db *sql.DB) ([]byte, error) {
 	buf.WriteString("\n--\n-- Dbmate schema migrations\n--\n\n")
 
 	if len(migrations) > 0 {
-		buf.WriteString("INSERT INTO schema_migrations (version) VALUES\n    (" +
+		buf.WriteString("INSERT INTO public.schema_migrations (version) VALUES\n    (" +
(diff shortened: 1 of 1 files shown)


pkqk:
Looks like this was caused by a change in [9.6.9](https://www.postgresql.org/docs/9.6/static/release-9-6-8.html)

> Avoid use of insecure search_path settings in pg_dump and other client programs (Noah Misch, Tom Lane)

> pg_dump, pg_upgrade, vacuumdb and other PostgreSQL-provided applications were themselves vulnerable to the type of hijacking described in the previous changelog entry; since these applications are commonly run by superusers, they present particularly attractive targets. To make them secure whether or not the installation as a whole has been secured, modify them to include only the pg_catalog schema in their search_path settings. Autovacuum worker processes now do the same, as well.

> In cases where user-provided functions are indirectly executed by these programs — for example, user-provided functions in index expressions — the tighter search_path may result in errors, which will need to be corrected by adjusting those user-provided functions to not assume anything about what search path they are invoked under. That has always been good practice, but now it will be necessary for correct behavior. (CVE-2018-1058)

krevels:
Currently encountering this issue as well. Would love to see it included.

PR #66, opened 2019-04-17
Add pending command to list pending migrations

matguig:


subhi-chegg:
This command will be very useful.
Is there a release scheduled for this?

amacneil:
Sorry for the delay, I've been thinking about the best way to integrate this with the other requests for a `dbmate status` command. I think we could solve both use cases with a single command.

Issue #46, opened 2018-06-06
Cannot run inside a transaction block


pauldraper:
PostreSQL

```
-- migrate:up
ALTER TYPE my_enum ADD VALUE 'my_value';
```

Running this migrations results in

```
Error: pq: ALTER TYPE ... ADD cannot run inside a transaction block
```

camerondavison:
is there a work around for this?

amacneil:
For this particular issue, you could try falling back to the old way of modifying an enum, it might work inside a transaction:

https://stackoverflow.com/a/10404041

More generally, it would be nice if there was a way to disable our transaction logic (run a specific migration without automatically wrapping in a transaction). Something like:

```
-- migrate:up transaction:false
```

I'm open to suggestions on the syntax and/or a PR implementing this, it should be quite easy.

adrianomedeirossantos:
Hi @amacneil, it would be great to have support for this feature.  I'm currently working in a project which needs to run migrations out of a transaction.  I need to run the following migration:

```
DROP INDEX CONCURRENTLY my_idx;
```

Unfortunately,  `CONCURRENTLY` cannot be executed within a transaction block as you can see in http://www.postgresqltutorial.com/postgresql-indexes/postgresql-drop-index/. 

Do you think this feature could be implemented any soon?

pauldraper:
> -- migrate:up transaction:false

Yes I agree this is right.

amacneil:
I think this feature would be pretty straight forward to implement - it just needs some bike shedding on the comment syntax, additional parsing logic to detect when transactions are disabled, and some light refactoring to not automatically open the transaction.

I don't currently have time to tackle it, but would happily review a PR.

benjreinhart:
Addressed in #68

PR #68, opened 2019-05-16
Add support for options; transaction option

benjreinhart:
Addresses #46 

This PR adds support for options. The only option implemented here is `transaction` which runs the the migration outside of a transaction when `false`. This is useful for certain statements which cannot run inside of a transaction, such as the `ALTER TYPE` statement used for adding values to enums in Postgres.

I took care to make sure this didn't break any existing functionality, but please point out if I missed something.

I'd love to end-to-end test this, but not every db in the test suite supports PG's enum syntax. I'm sure we could restructure some of the tests such that we end-to-end test this, but I only have so much time I can spend on this. For now, it's unit tested and should cover all the necessary cases.

### DIFF
--- a/pkg/dbmate/migrations.go
+++ b/pkg/dbmate/migrations.go
@@ -1,0 +1,133 @@
+package dbmate
+
+import (
+	"io/ioutil"
+	"regexp"
+	"strings"
+)
+
+// MigrateOptions is an interface for accessing migration options
+type MigrateOptions interface {
+	SkipTransaction() bool
+}
+
+type migrateOptions map[string]string
+
+// SkipTransaction returns true if the migration is to run outside a transaction
+// Defaults to false.
+func (m migrateOptions) SkipTransaction() bool {
+	return m["skip_transaction"] == "true"
+}
+
+// NewMigrateOptions returns a MigrateOptions interface
+func NewMigrateOptions() MigrateOptions {
+	return make(migrateOptions)
+}
+
+// Migrate contains 'up' or 'down' migration commands and options
+type Migrate struct {
+	Direction string
+	Contents  string
+	Options   MigrateOptions
+}
+
+// NewMigrate constructs a Migrate object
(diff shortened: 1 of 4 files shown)


pkg/dbmate/migrations_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	up, down = parseMigrationContents(migration)
+
+	require.Equal(t, "-- migrate:up\ncreate table users (id serial, name text);", up.Contents)
+	require.Equal(t, false, up.Options.SkipTransaction())
+
+	require.Equal(t, "-- migrate:down\ndrop table users;", down.Contents)
+	require.Equal(t, false, down.Options.SkipTransaction())
+
+	migration = `-- migrate:up skip_transaction:true
```
amacneil: maybe add a comment here saying that the reason you chose this particular operation is that it will fail inside a transaction

pkg/dbmate/migrations.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+var whitespaceRegExp = regexp.MustCompile(`\s+`)
+var optionSeparatorRegExp = regexp.MustCompile(`:`)
+
+// parseMigrationOptions parses the options portion of a migration
+// block into an object that satisfies the MigrationOptions interface,
+// i.e., the 'transaction:false' piece of the following:
+//
+//     -- migrate:up transaction:false
```
amacneil: note that your example here does not match the `skip_transaction:true` format you actually implemented. although I have a slight preference for `transaction:false` syntax to avoid the double negative.

benjreinhart: whoops.. yeah I will change to `transaction:false` anyways

pkg/dbmate/migrations_test.go
benjreinhart: can do

PR #76, opened 2019-05-24
Fix special chars passwords in MySQL Driver

### LINKED ISSUE #57: DB connection strings do not support exclamation points in password
I was getting an "Error 1045: Access denied for user...", given a valid connection string. Checked privileges and made sure everything was right on database side. It seems that mysql is fine with connection strings with exclamation points in them, while net/url mangles them and passes them to mysql as %21, which mysql interprets literally.

kriive:
This PR aims to fix #57 
When MySQL password included special chars such as exclamation point or @ (at) MySQL backend errored out (invalid password).

https://github.com/amacneil/dbmate/blob/87c22515a9511c680e0391199cc3f53e0fc57eaf/pkg/dbmate/mysql.go#L21-L39

url.userinfo.String() (which gets called inside url.String()) returns %-encoded strings and MySQL interprets it as an actual password.
Now the function percent-decodes it first before returning.

This PR adds a test to cover special chars in MySQL.

### DIFF
--- a/pkg/dbmate/mysql.go
+++ b/pkg/dbmate/mysql.go
@@ -19,25 +19,25 @@ type MySQLDriver struct {
 }
 
 func normalizeMySQLURL(u *url.URL) string {
-	// set default port
-	host := u.Host
+	temporaryURL := *u // make a copy of u
 
-	if u.Port() == "" {
+	// set default port
+	host := temporaryURL.Host
+
+	if temporaryURL.Port() == "" {
 		host = fmt.Sprintf("%s:3306", host)
 	}
 
 	// host format required by go-sql-driver/mysql
 	host = fmt.Sprintf("tcp(%s)", host)
 
-	query := u.Query()
+	query := temporaryURL.Query()
 	query.Set("multiStatements", "true")
 
-	// I reutilized u, is this good practice?
-	// Should we make a copy of u and then work on it instead?
-	u.RawQuery = query.Encode()
+	temporaryURL.RawQuery = query.Encode()
 
 	// Get decoded user:pass
-	userPassEncoded := u.User.String()
+	userPassEncoded := temporaryURL.User.String()
 	userPass, _ := url.QueryUnescape(userPassEncoded)
 
 	// Build DSN w/ user:pass percent-decoded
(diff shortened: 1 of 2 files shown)


pkg/dbmate/mysql.go
```diff
@@ -35,7 +35,9 @@ func normalizeMySQLURL(u *url.URL) string {
 	normalizedURL.RawQuery = query.Encode()
 
 	str := normalizedURL.String()
-	return strings.TrimLeft(str, "/")
+	decodedValue, _ := url.QueryUnescape(str)
```
amacneil: need to handle this error value somehow

kriive:
Sure thing! I'll look into it asap. 
How do you think we should handle user:pass QueryUnescape err?

benjreinhart:
@kriive I think we can use `url.User.Username()` and `url.User.Password()` to get the unescaped versions and then compose the url manually. I think something like the following works

```go
func normalizeMySQLURL(u *url.URL) string {
	username := u.User.Username()
	password, _ := u.User.Password()

	host := u.Host

	// set default port
	if u.Port() == "" {
		host = fmt.Sprintf("%s:3306", host)
	}

	// host format required by go-sql-driver/mysql
	host = fmt.Sprintf("tcp(%s)", host)

	path := u.Path

	query := u.Query()
	query.Set("multiStatements", "true")
	queryString := query.Encode()

	return fmt.Sprintf("%s:%s@%s%s?%s", username, password, host, path, queryString)
}
```

kriive:
@benjreinhart thank you very much for your suggestion! If we decide to use url.User.Username() and url.User.Username() we should also test whether user and/or password are blank.

I think url.User.String() is a better choice (https://golang.org/src/net/url/url.go?s=5107:5151#L419), what do you think? We only need to unescape the result ourselves.

pkg/dbmate/mysql.go
```diff
@@ -19,23 +19,38 @@ type MySQLDriver struct {
 }
 
 func normalizeMySQLURL(u *url.URL) string {
-	normalizedURL := *u
-	normalizedURL.Scheme = ""
+	temporaryURL := *u // make a copy of u
```
benjreinhart: Nitpick, but we don't need to do this if you don't modify the url, i.e., remove the `temporaryURL.RawQuery = query.Encode()` piece and instead store `query.Encode()` in a new variable. I'd vote for using variables instead of constructing the url in this particular case.

PR #80, opened 2019-06-06
Provide a reasonable error message to users

schancel:
Currently, when the .env file cannot be parsed, the dbmate application
bombs with no explanation.  This commit prints out the error.

### DIFF
--- a/main.go
+++ b/main.go
@@ -123,7 +123,7 @@ func loadDotEnv() {
 	}
 
 	if err := godotenv.Load(); err != nil {
-		log.Fatal("Error loading .env file")
+		log.Fatalf("Error loading .env file: %s", err.Error())
 	}
 }
 


main.go
```diff
@@ -123,7 +123,7 @@ func loadDotEnv() {
 	}
 
 	if err := godotenv.Load(); err != nil {
-		log.Fatal("Error loading .env file")
+		log.Fatalf("Error loading .env file: %q", err.Error())
```
amacneil: How does this print? I think just `%s` would be sufficient?

schancel: Yes.  It prints with quotes around it currently.

amacneil: Can you remove the quotes to make consistent with our other error messages?

Also - do you have an example of an error this would print?

schancel: In my case it was erroring as such:

```
2019/06/13 19:51:46 Error loading .env file: Can't separate key from value
```

PR #84, opened 2019-07-22
Add dry-run feature to comparison

emmetog:


amacneil:
Closing stale PR - doesn't seem sufficient enough of a feature to call out in the alternatives matrix

PR #94, opened 2019-09-26
clarify that it simply uses SQL to define migrations

AndreaCrotti:
Not sure if my corrections is good, but as it is now it seems to suggest that there is a custom built DSL, while the DSL is just SQL, which I think everyone looking at a database migration tool is quite familiar with anyway.


amacneil:
Oh yeah - at the time I wrote this it was intended to be a joke, but happy to clarify this!

PR #115, opened 2020-01-20
Add support for repeatable SQL scripts

ncwade:
Howdy! I'm looking to add a feature similar to Flyway's repeatable migrations. We tend to stored our stored procedures/functions as scripts like so;
```sql
CREATE OR REPLACE FUNCTION add_user (integer, text)
RETURNS integer AS $total$
DECLARE
	total integer;
BEGIN
    insert into users (id, name) values ($1, $2);
    select count(id) into total from users;
    return total;
END;
$total$ LANGUAGE plpgsql;
```

I added it in under the term "repeatable", and it works by storing the checksum of the last script loaded. This new feature would allow us to simply run `dbmate up` and have our schema and stored procedures all loaded and ready to go. You simply take the previous example and annotate it like so;
```sql
-- migrate:repeatable
CREATE OR REPLACE FUNCTION add_user (integer, text)
RETURNS integer AS $total$
DECLARE
	total integer;
BEGIN
    insert into users (id, name) values ($1, $2);
    select count(id) into total from users;
    return total;
END;
$total$ LANGUAGE plpgsql;
```

If this is a feature you are interested in pulling into the main I'd love some feedback!


amacneil:
Thanks for the PR! Can you explain the feature a little more? Sorry I'm not familiar with flyway. Is this basically some sort of "pre-migration" script that runs every single time you `dbmate up`?

ncwade:
Not every time, just when a the script changes. The general idea is that instead of having a version(like a migration), some objects just have a checksum. The entire script would be (re-)applied whenever the checksum of the file changes. This useful for objects who can be clearly defined in a single file(views/stored procedures/functions). Since they are defined as dedicated scripts in git, the differences can be easily seen in pull requests.

amacneil:
Closing stale PR since it has a lot of conflicts.

This feature adds a lot of complexity to dbmate (new migration directories, new migrations table, new commands) and it seems to be rather niche (no other comments on this issue, not something I've personally needed in the past). If it does turn out to be a popular request, I also think there are probably simpler ways to solve it (e.g. just create a new migration each time you update the function, or alternatively some sort of include directive in migrations would allow you to run it each time you need to use the function without worrying about if the function has changed).

PR #120, opened 2020-02-28
Add status command

fernandocarletti:
Hello!

I tried to finish the work for a status command started by @hmleal on PR #32.

It wasn't much easy to unit test it, since most of it is about the presentation. I managed to separate a bit the logic from the presentation and at least test the data extracted from the current migrated versions.

I also tried to implement the "summary only" as suggested by @MarkMurphy, but I could not quickly find a way to add a command specific flag, should be a quick in once it is figured out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ dbmate rollback  # roll back the most recent migration
 dbmate down      # alias for rollback
 dbmate dump      # write the database schema.sql file
 dbmate wait      # wait for the database server to become available
+dbmate status    # show the status of all migrations
 ```
 
 ## Usage
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,3 +18,5 @@ services:
 
   postgres:
     image: postgres:9.6
+    environment:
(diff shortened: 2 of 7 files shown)


fernandocarletti:
@amacneil The commit https://github.com/amacneil/dbmate/pull/120/commits/d8a087f862cce23de1a5fb971fcb747273814833 fixes the following error
https://travis-ci.org/amacneil/dbmate/builds/656116207#L559. I had it locally as well, let me know if this should be handled differently.

amacneil:
Thanks - I just merged the postgres docker fix separately in https://github.com/amacneil/dbmate/pull/121.

Will take a look at this this week hopefully. It looks good, although I do like the idea @MarkMurphy suggested of existing with non-zero code if there are pending migrations (useful for scripts etc).

fernandocarletti:
I think I missed that part in the comment, it does sounds good. But should we add a flag (`--error-on-pending` or something)? Having pending migrations is not really an error, it is actually expected in some use cases. WDYT?

Issue #101, opened 2019-11-04
Exclusive lock during migration


marton78:
Hi,

as far as I can tell, currently there is now explicit lock applied during migrations. That would mean that it is dangerous to run migrations in a concurrent setting, e.g. when several instances of some API server including dbmate are running.

It would be advisable to obtain an exclusive lock of the `migrations` table before migrating.

amacneil:
This is a good feature request, I've discussed it with others in person too. Happy to accept a PR adding this.

marton78:
Sorry but I need to pass, I have zero knowledge of Golang :)

amacneil:
No problem, I think someone else is already working on this. If not I will keep on the backlog.

palmenhq:
I miss this also. Is somebody working on it already? Otherwise I’d be happy to give it a go (pun intended, hehe). If so, I see a few paths that could be taken, and I’d just like to check what’s your opinion before moving forward :)

Firstly, do you think this would be better as a default or an opt-in? I’m thinking of something like a `--lock` or `--concurrent` flag. To me making it default would make sense though as it _shouldn't_ really affect backwards compatibility of usage, right? And I don't really see a use case where you'd want to opt out or? Possibly to avoid deadlocks but.. idk, sounds unlikely. Maybe an opt-out flag would be better for that case?

An approach I'm thinking of is a two-phase locking based on hostname where one host acquires the lock, runs pending migrations and releases the lock again - so the next host can do the same after waiting for the first lock to release. In the case of a deadlock this makes it easy to introspect the database and find which host caused it, as opposed to using something like one-time generated ids. Possibly one could make a hybrid, for the case of someone running concurrent migrations on the same host? Or maybe that's just over engineering it?

Finally, I'm thinking of how to solve testing. This feature would need two dbmate implementations to run synchronised if it should be tested againts a real database i guess? So maybe we could add another testing phase in docker, that tests this particular feature? Additionally to some unit tests. Whats your thoughts on this?

Cheers!

amacneil:
Hey, contribution sounds like a great idea!

I think it would be fine to make this behavior the default. I can't even think of a good reason to support opt-out: it should be completely transparent to people unless you happen to try to run multiple dbmates at the same time. Let me know if you can think of a good reason to expose this to users, but my initial reaction is that it should just be default behavior to prevent concurrent migrations.

[...]

PR #125, opened 2020-03-20
Exclusive migration locks

palmenhq:
A first take on addressing https://github.com/amacneil/dbmate/issues/101

I have a couple of minor issues idk how to tackle, so would be amaze with some feedback :)

One small thing: Do you think we should have a timeout on the change lock (in case of deadlocks) or nah? I think not having it would be better, considering some migrations could take a lot of time and the risk for deadocks is quite low considering the locking expires on session termination.

The main thing: Idk exactly what to do about SQLite, shall we skip it or implement our own locking mechanism? Skipping _could_ be OK for SQLite in particular I guess, because I don't really see a use case where you run concurrent migration scripts against a single database file. That being said, it would also feel bad to skip one particular driver, but due to SQLite not having any locking mechanism for a full open session (that I can find, that is) it becomes slightly more complicated than for the other drivers. Totally solvable though, so just let me know your thoughts and I'll make some final fixes accordingly!

### DIFF
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -283,7 +283,7 @@ func (db *DB) Migrate() error {
 	}
 	defer mustClose(sqlDB)
 
-	err = waitForOthersChangeLockToReleaseAndAcquireChangeLock(drv, sqlDB)
+	err = waitForOtherLockToReleaseAndAcquireMigrationLock(drv, sqlDB)
 	if err != nil {
 		return err
 	}
@@ -335,7 +335,7 @@ func (db *DB) Migrate() error {
 		_ = db.DumpSchema()
 	}
 
-	err = drv.ReleaseChangeLock(sqlDB)
+	err = drv.ReleaseMigrationLock(sqlDB)
 	if err != nil {
 		return err
 	}
@@ -463,20 +463,20 @@ func (db *DB) Rollback() error {
 	return nil
 }
 
-func waitForOthersChangeLockToReleaseAndAcquireChangeLock(drv Driver, sqlDB *sql.DB) error {
+func waitForOtherLockToReleaseAndAcquireMigrationLock(drv Driver, sqlDB *sql.DB) error {
 	counter := 0
-	for { // FIXME should we have a timeout for change locks? Or no?
-		acquiredChangeLock, err := drv.AcquireChangeLock(sqlDB)
+	for {
+		successfullyAcquiredLock, err := drv.AcquireMigrationLock(sqlDB)
 		if err != nil {
 			return err
 		}
 
-		if acquiredChangeLock {
+		if successfullyAcquiredLock {
 			break
 		}
 
 		if counter == 0 {
-			fmt.Print("Waiting other for database change lock to release")
+			fmt.Print("Waiting other for database migration lock to release")
 		} else {
 			fmt.Print(".")
 		}
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -283,6 +283,11 @@ func (db *DB) Migrate() error {
 	}
 	defer mustClose(sqlDB)
 
+	err = waitForOthersChangeLockToReleaseAndAcquireChangeLock(drv, sqlDB)
+	if err != nil {
+		return err
+	}
+
 	applied, err := drv.SelectMigrations(sqlDB, -1)
 	if err != nil {
 		return err
@@ -328,6 +333,11 @@ func (db *DB) Migrate() error {
 	// automatically update schema file, silence errors
 	if db.AutoDumpSchema {
 		_ = db.DumpSchema()
+	}
+
+	err = drv.ReleaseChangeLock(sqlDB)
+	if err != nil {
+		return err
 	}
 
 	return nil
@@ -450,6 +460,32 @@ func (db *DB) Rollback() error {
 		_ = db.DumpSchema()
 	}
 
+	return nil
+}
+
+func waitForOthersChangeLockToReleaseAndAcquireChangeLock(drv Driver, sqlDB *sql.DB) error {
+	counter := 0
+	for { // FIXME should we have a timeout for change locks? Or no?
+		acquiredChangeLock, err := drv.AcquireChangeLock(sqlDB)
+		if err != nil {
+			return err
+		}
+
+		if acquiredChangeLock {
+			break
+		}
+
+		if counter == 0 {
+			fmt.Print("Waiting other for database change lock to release")
+		} else {
+			fmt.Print(".")
+		}
+		time.Sleep(time.Second)
+		counter += 1
+	}
+	if counter > 0 {
+		fmt.Print("\n")
+	}
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/dbmate/driver.go
```diff
@@ -17,6 +17,8 @@ type Driver interface {
 	SelectMigrations(*sql.DB, int) (map[string]bool, error)
 	InsertMigration(Transaction, string) error
 	DeleteMigration(Transaction, string) error
+	AcquireChangeLock(*sql.DB) (bool, error)
```
amacneil: This is a nit but I would probably call these methods `AcquireMigrationLock` and `ReleaseMigrationLock`, or maybe just `AcquireLock`? ChangeLock sounds a little weird to me.

pkg/dbmate/db.go
```diff
@@ -330,6 +335,11 @@ func (db *DB) Migrate() error {
 		_ = db.DumpSchema()
 	}
 
+	err = drv.ReleaseChangeLock(sqlDB)
```
amacneil: this should probably live inside a `defer` block

pkg/dbmate/mysql.go
```diff
@@ -239,6 +239,29 @@ func (drv MySQLDriver) DeleteMigration(db Transaction, version string) error {
 	return err
 }
 
+// Acquires a change lock by setting an
+// [exclusive lock](https://dev.mysql.com/doc/refman/5.7/en/locking-functions.html#function_get-lock). NB! Locks are
+// reference counted and hence you must call `ReleaseChangeLock` the same number of times you call `AcquireChangeLock`
+// successfully
```
amacneil: what happens if the dbmate process crashes and doesn't release the lock?

pkg/dbmate/postgres.go
```diff
@@ -174,6 +174,33 @@ func (drv PostgresDriver) DeleteMigration(db Transaction, version string) error
 	return err
 }
 
+// The sum of all charCodes in the string "dbmate" :)
+const advisoryLockKey = 621
```
amacneil: hmm, this could cause weird behavior for some applications if we don't have any way to namespace it...

pkg/dbmate/sqlite.go
```diff
@@ -165,6 +165,18 @@ func (drv SQLiteDriver) DeleteMigration(db Transaction, version string) error {
 	return err
 }
 
+// Does nothing currently
+func (drv SQLiteDriver) AcquireChangeLock(db *sql.DB) (bool, error) {
+	// FIXME use change locking for this driver also?
+	fmt.Println("Skipping migration locking because database driver is SQLite")
```
amacneil: I haven't done much research on this, but it looks like sqlite has something called an exclusive lock for the file - I wonder if we could use that.

pkg/dbmate/db.go
```diff
@@ -453,6 +463,32 @@ func (db *DB) Rollback() error {
 	return nil
 }
 
+func waitForOthersChangeLockToReleaseAndAcquireChangeLock(drv Driver, sqlDB *sql.DB) error {
+	counter := 0
+	for { // FIXME should we have a timeout for change locks? Or no?
```
amacneil: rather than manually implementing the loop, could you instead just use the non-try versions of the locking functions? e.g. `pg_advisory_lock` instead of `pg_try_advisory_lock`

that way it would just automatically block until the lock is available. and it would save you manually worrying about the timeout.

pkg/dbmate/db.go
```diff
@@ -453,6 +463,32 @@ func (db *DB) Rollback() error {
...
+		if acquiredChangeLock {
+			break
+		}
+
+		if counter == 0 {
+			fmt.Print("Waiting other for database change lock to release")
+		} else {
+			fmt.Print(".")
+		}
+		time.Sleep(time.Second)
```
amacneil: if we do keep this loop here you need to make sure that ctrl+c is handled correctly

amacneil:
Sweet, thanks! Apologies in advance that I'm usually pretty slow with code reviews.

I'm not sure about the timeout. It seems simpler without it. I'm thinking of a situation where someone is running dbmate before their application boots, e.g. `dbmate up && node app.js`. In this case they would not want the app to start until migration is complete. On the other hand, if the migration hangs for some reason, it would mean that the app never starts.

Sqlite: I added one suggestion above (exclusive lock - `BEGIN EXCLUSIVE TRANSACTION`), although I haven't researched it much. I think it might be a problem that they don't support nested transactions, and we do all the migrations inside a transaction. So you would need to refactor the migration code so that it uses named save points rather than opening and closing a separate transaction for each migration (assuming that is possible).

pkg/dbmate/mysql.go
palmenhq: Its released on session termination, but probably that should be clarified.

pkg/dbmate/postgres.go
palmenhq: Yes this is true, I'm unsure how to tackle that problem, its hard to namespace since [postgres only accepts numbers](https://www.postgresql.org/docs/9.6/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS) as far as i can see. We could either set this to something bigger and more random i guess, but it still feels shaky. Or we could namespace it further by setting two numbers, which effectively gives the same result, but assumingly with less collision risk? Maybe if this happens to collide with an application's lock key we can always give the user the option to specify it themselves - on the other hand as a user tracking that down as the solution to such a problem seems... hard. Idk, how much of a dealbreaker is this do you think?

pkg/dbmate/db.go
palmenhq: Sure! This was solely for the the ticking `...` effect, which in a way I like. For me it would make it easier to understand how long the wait has been, if something is long-running.

pkg/dbmate/db.go
palmenhq: Not sure i understand, a `time.Sleep` would terminate itself on ctrl+c no?

palmenhq:
Regarding the timeout, coming back to it I think it doesn't make much sense. In the example of `dbmate up && node app.js` a timeout should probably rather crash `dbmate up`, meaning `node app.js` is not run anyways. Since the current implementation, as far as i can tell, has a very low risk of deadlocks (considering the locks are released on session termination) imo it should be fine.

As for Sqlite, after some more digging I think we could use [`PRAGMA locking_mode = EXCLUSIVE`](https://www.sqlite.org/pragma.html#pragma_locking_mode) at the start of the migration. The only problem I can see with this (the same problem is true for `BEGIN EXCLUSIVE TRANSACTION` though) is that all other interactions (also reads) with the database will fail with `Error: database is locked` while migrations are in progress. Unsure if this is a desired behaviour really, considering that if you have application A and B working against the same database the reason is probably that you want application B to function while application A is running migrations, right?

pkg/dbmate/db.go
palmenhq: How do i do that while still checking for errors though? 😬 panicking seems a little dramatic

amacneil:
Closing this PR since it is stale and has a lot of conflicts.

I still think this is a useful feature, so if someone would like to pick it up in the future I'm happy to merge it.

PR #129, opened 2020-04-27
Oracle Support for DBMate

csuriano23:
This PR adds Oracle support to DBMate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .env
+.idea
 /dbmate
 /dist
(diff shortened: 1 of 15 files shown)


amacneil:
Closing stale PR.

I'm open to adding Oracle support but I am not personally as familiar with it. It appears that this PR adds a lot of complexity to the build and Dockerfile which is not ideal.

The "dbmate engine" implemented here (for parsing sql statments) is also something which could be useful for other drivers, but would be best in a separate PR.

PR #131, opened 2020-04-30
Add MySQL connection through socket file

csuriano23:
For MySQL connections, if a `socket` parameter is specified in query URL the connection to the database goes through the unix socket file specified


amacneil:
Thanks!

PR #138, opened 2020-05-28
adding verbose output for statement execution

vperiyasamy:
Adding a verbose command line flag to allow users to print the result of each statement execution. Should not disrupt the existing flow for users as it's completely optional. Summary of changes:
- new `verbose` flag
- new field for `db` to print result of migrations
- new interface methods for drivers to pass verbose flag
- capture and print of result for every `exec` based on verbose flag across db and drivers

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ The following command line options are available with all commands. You must use
 * `--schema-file, -s "./db/schema.sql"` - a path to keep the schema.sql file.
 * `--no-dump-schema` - don't auto-update the schema.sql file on migrate/rollback
 * `--wait` - wait for the db to become available before executing the subsequent command
-* `--verbose, -v` - print the result of each statement execution
+* `--print-result` - print the result of each statement execution
 
 For example, before running your test suite, you may wish to drop and recreate the test database. One easy way to do this is to store your test database connection URL in the `TEST_DATABASE_URL` environment variable:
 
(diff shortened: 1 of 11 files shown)


vperiyasamy:
changed name from `--verbose` because urfave cli was panicking at redefining the flag `v`. new flag is now called `print-result`

vperiyasamy:
@amacneil - not sure how to assign reviewers

amacneil:
Hey, sorry for the delay on this, it has been a busy few days.

Some quick thoughts/feedback:
* I think it would make more sense for the flag to exist on the command, rather than globally (it's a `urfave/cli` thing). So this would make the argument `dbmate up --verbose` rather than `dbmate --verbose up`. I assume that might also fix your problem with overloading the `-v` shorthand.
* Any thoughts on how to test this?
* What does the output look like? Is it worth formatting that `result` struct to make it prettier?

Otherwise looks good, thanks for the contribution!

vperiyasamy:
thanks for the comments, no worries. I think I've addressed these concerns with the new commit:
- moved flag to subcommands for `up`, `migrate`, and `rollback` - you were right, I am able to use verbose now
- I added a test in `db_test.go` which captures the stdout and makes sure it reports the `Rows affected` value
- output is a formatted string I've created, you can see it in `utils.go`
- originally in this PR I had added verbose statements to all of the driver execution results, but after testing I found out that these have no effect because there is no `InsertId` or `RowsAffected` for those commands, so have removed those 👍

vperiyasamy:
@amacneil bump

vperiyasamy:
@amacneil sorry for the pestering, but any chance you would be able to look at this soon? we feel that without this logging we are missing a crucial sanity check when running our db migrations at scale

amacneil:
Cool, this looks good. Sorry I have been busy lately and not much time for OSS. I added you as a contributor, so if you're happy with it and it's passing CI feel free to merge!

vperiyasamy:
@amacneil thank you! no worries, I appreciate it.

PR #140, opened 2020-06-10
Add ClickHouse support

Flamefork:


Flamefork:
Rebased after 1.9 release

Makefile
```diff
@@ -52,4 +53,5 @@ docker-all:
 
 .PHONY: docker-bash
 docker-bash:
-	-docker-compose run --rm dbmate bash
+	docker-compose build
+	docker-compose run --rm dbmate bash
```
amacneil, line 57: can you revert this bit

go.sum
```diff
@@ -1,12 +1,20 @@
...
+github.com/bkaradzic/go-lz4 v1.0.0/go.mod h1:0YdlkowM3VswSROI7qDxhRvJ3sLhlFrRRwjwegp5jy4=
+github.com/cloudflare/golz4 v0.0.0-20150217214814-ef862a3cdc58 h1:F1EaeKL/ta07PY/k9Os/UFtwERei2/XzGemhpGnBKNg=
+github.com/cloudflare/golz4 v0.0.0-20150217214814-ef862a3cdc58/go.mod h1:EOBUe0h4xcZ5GoxqC5SDxFQ8gwyZPKQoEzownBlhI80=
 github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/cpuguy83/go-md2man/v2 v2.0.0 h1:EoUDS0afbrsXAZ9YQ9jdu/mZ2sXgT1/2yyNng4PGlyM=
 github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/go-sql-driver/mysql v1.4.0/go.mod h1:zAC/RDZ24gD3HViQzih4MyKcchzm+sOG5ZlKdlhCg5w=
```
amacneil, line 14: I think you need to run `go mod tidy` to clean up these

pkg/dbmate/clickhouse.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// CreateMigrationsTable creates the schema_migrations table
+func (drv ClickHouseDriver) CreateMigrationsTable(db *sql.DB) error {
+	_, err := db.Exec(`
+		create table if not exists schema_migrations (
+			version String,
+			ts DateTime default now(),
+			applied UInt8 default 1
+		) engine = ReplacingMergeTree(ts)
```
amacneil, line 184: can you keep this consistent with the table we create for other drivers. we might add an applied timestamp at some point, but I'd like to be consistent and do it for all drivers.

Makefile
Flamefork: Reverted. Sorry, it must be an artefact of a recent rebase.

go.sum
Flamefork: I did this already, doublechecked now with the same effect. Not sure why is it adding these lines.

Flamefork: But, I've just done it completely from scratch and it worked.

pkg/dbmate/clickhouse.go
Flamefork: The reason this column (and a ts column) exist is that ClickHouse is append-only DB. There is no usable way to update or delete a row. So, in order to be able to mark a migration as rolled back, we need to insert a new row with the same version but applied = false and newer ts. This is an idiomatic way to do updates in ClickHouse.

amacneil: Makes sense, thanks

README.md
```diff
@@ -161,6 +161,12 @@ To specify an absolute path, add an additional forward slash to the path. The fo
 DATABASE_URL="sqlite:////tmp/database_name.sqlite3"
 ```
 
+**ClickHouse**
+
+```sh
+DATABASE_URL="clickhouse://127.0.0.1:9000?username=username&password=password&database=database_name"
```
amacneil, line 167: It's a bit silly that the clickhouse-go driver doesn't use the username/password fields already supported in a standard URL format (`username:password@127.0.0.1:9000`). Would be nice to support these as well in order to be consistent with other dbmate drivers.

amacneil: Actually, it's also weird that they expect the `database` as a URL parameter. The official [clickhouse jdbc driver](https://github.com/ClickHouse/clickhouse-jdbc) even uses a more standard format.

Could you add a bit of extra URL parsing (similar to what I do in the mysql driver) to support a standard url format like:

```
clickhouse://username:password@host:port/dabase_name
```

Under the hood you can convert these to the format `clickhouse-go` expects, and you could still support these query parameters if they are passed (just leave them unchanged) for interoperability? That will allow us to follow dbmate standards while also interop with clickhouse-go.

README.md
```diff
@@ -343,10 +349,11 @@ Why another database schema migration tool? Dbmate was inspired by many other to
 |Automatically load .env file||||||:white_check_mark:|
 |No separate configuration file||||:white_check_mark:|:white_check_mark:|:white_check_mark:|
 |Language/framework independent|:eight_pointed_black_star:|:eight_pointed_black_star:|:eight_pointed_black_star:|||:white_check_mark:|
-| **Drivers** |||||||
+| **Drivers** |
 |PostgreSQL|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|
 |MySQL|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|
 |SQLite|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|
+|ClikHouse|||:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|
```
amacneil, line 356: typo in `ClickHouse`

amacneil:
Thanks - I added one more comment on the URL format, and a small typo. After that should be good to merge.

README.md
Flamefork: Added parsing and a link to other CH-related connection string options.

Flamefork:
😇

Flamefork:
Thank you!

Issue #110, opened 2019-12-25
PostrgeSQL schema setting for migrations table


maxtryk:
Hi Adrian!

Is there a way to set the name of the postgres schema in which `schema_migrations` is created?

amacneil:
Hi! Not currently, we default to using the `public` schema. 

What's your use case? E.g. are you storing independent dbmate-managed databases in different schemas? Or do you just choose to store everything in a schema with a non-default name? If you were going to customize the schema name for dbmate where would you expect to set it?

maxtryk:
>Or do you just choose to store everything in a schema with a non-default name?

Correct.

>If you were going to customize the schema name for dbmate where would you expect to set it?

It feels like a command line option would do it - this stuff shouldn't change between environments...

Thank you for your reply and happy holidays!

Rested:
Would you want it to create the schema (if it does not exist) or not?

Videl:
Seems the schema is created if needed when using this:
``
DATABASE_URL="postgres://postgres:password@db:5432/db?sslmode=disable&search_path=myschema"
``

But the ``schema_migrations`` is still in public. Maybe it could follow the schema setting?

enverbisevac:
This is very useful in multitenant scenarios for Semi-isolated approach.

RabidFire:
Hi @amacneil - First off, love the vision of this library. Appreciate all the hard work! We currently use golang-migrate, but want to introduce non-Go services and were looking for something more agnostic. 😄 

Our team needs this feature to use this library, and I'd be happy to send in a PR if you could provide some guidance and verify these steps:

### Use case
Our team has a bunch of microservices, each talking to a different schema within the same database. Each microservice manages its own migrations table. This allows separation of concerns and running migrations independently.

### Initial Research
Did some digging and saw that `public.schema_migrations` is [hard-coded](https://github.com/amacneil/dbmate/blob/ed9e57a4ade7a83584be46de364e6bcff1a5f611/pkg/dbmate/postgres.go#L168). In particular, it's only hard-coded for Postgres and no other database.

My first thought was to remove `public.` from `public.schema_migrations` everywhere so that it would work automatically when someone passes in a `search_path` parameter within `DATABASE_URL` (like @Videl [commented](https://github.com/amacneil/dbmate/issues/110#issuecomment-665915427)). However, I see this was introduced by #49 because of the way `pg_dump` works. 

### Possible Solutions
[...]

PR #166, opened 2020-10-31
postgres: Limit pg_dump to schemas in search_path

amacneil:
Implementing suggestion from @RabidFire in [this comment](https://github.com/amacneil/dbmate/issues/110#issuecomment-711141575).

If `search_path` is specified in a postgres connection URL, the dump command currently fails (because the `search_path` parameter is custom to go's `lib/pq` driver):

```
$ dbmate dump
Error: pg_dump: [archiver (db)] connection to database "" failed: invalid URI query parameter: "search_path"
```

In this PR we update the dump command to explicitly export only schemas specified in `search_path`.

### DIFF
--- a/pkg/dbmate/postgres.go
+++ b/pkg/dbmate/postgres.go
@@ -19,7 +19,7 @@ func init() {
 type PostgresDriver struct {
 }
 
-func normalizePostgresURL(u *url.URL) string {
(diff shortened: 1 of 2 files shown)


RabidFire:
@amacneil This is perfect! Thank you.

Issue #85, opened 2019-07-31
On MySQL 5.6 with default utf8mb4 encoding, schema_migrations table fails to create


charlietran:
Running dbmate on my MySQL 5.6 server with the DEFAULT CHARSET utf8mb4 gave me this error:

```
Error: Error 1071: Specified key was too long; max key length is 767 bytes
```
It's because on MySql 5.6 InnoDB with utb8mb4 encoding, the max length limit for any key is 191 characters (utf8mb4 chars are 4 bytes each). Digging into the dbmate code, it looks like it's because the `schema_migrations`table is created as such:

```golang
// CreateMigrationsTable creates the schema_migrations table
func (drv MySQLDriver) CreateMigrationsTable(db *sql.DB) error {
	_, err := db.Exec("create table if not exists schema_migrations " +
		"(version varchar(255) primary key)")

	return err
}
```

255 chars * 4 bytes = 1020 bytes, which will go over the limit. This command will fail even if the `schema_migrations` table has already been manually created with a lower varchar limit. 

To fix this, we could possibly:

* Detect if the version is 5.6 and limit the varchar to 191
* Specify latin1 encoding for the schema_migrations table, which only takes 1 byte per character

What do you think? Thanks for making a great tool!

andypalmer:
You can also set the mysql table defaults to work with longer keys:
```
set @@global.innodb_default_row_format=DYNAMIC;
set @@global.innodb_file_format = BARRACUDA;
set @@global.innodb_large_prefix = 1;
```

amacneil:
I tried to reproduce this issue but can't unfortunately. I tried using MySQL 5.6, manually adding `DEFAULT CHARSET=utf8mb4` to the create table statement, adding `?charset=utf8mb4` to the connection string, and none of them caused an issue when creating this table or inserting a migration.

I think given that the date stamps we store are max 14 characters, I'm happy limiting this column (or explicitly specifying `utf8` or `latin1` when creating the table) if that helps, but I'm worried about publishing a fix for this if I can't reproduce it locally.

Are you able to create a docker image reproducing the issue?

andypalmer:
I put together a [quick demo](https://github.com/andypalmer/dbmate_key_length_problem).

Enrico204:
Same here - version `5.5.5-10.1.41-MariaDB-0+deb9u1`. Apparently the max size was increased in MySQL 5.6 onwards, so this behavior is not reproducible with newer engines.

[...]

PR #172, opened 2020-11-01
Create the schema_migrations table using latin1 charset in MySQL

Enrico204:
This pull request force the `schema_migrations` table to be created with `latin1` character set, to fix the behavior of `dbmate` in MariaDB 10.1 and older (or MySQL 5.5 and older) when the system charset has more than 1 byte per character (e.g. utf8mb4) - see #85


amacneil:
Thanks!

For historical reference, I managed to reproduce the issue and test the PR by updating the mysql version in `docker-compose.yml` and forcing `utf8mb4` charset:

```
  mysql:
    image: mysql:5.6
    command: --character-set-server=utf8mb4
    environment:
      MYSQL_ROOT_PASSWORD: root
```

Issue #85 would not occur when using mysql 5.7 or later, or on 5.6 without explicitly setting `utf8mb4` charset.

PR #194, opened 2021-02-16
Support embedding the migrations into the binary with go1.16 io/fs.FS

### LINKED ISSUE #193: Support go1.16 io/fs.FS concept for file embedding
Hi there, it would be cool if the dbmate.DB type supported an FS field, so it can be used with go 1.16 file embedding. My use case is that I want to embed dbmate into my go program, and also embed the migrations using the `//go:embed` directive so I don't have to ship the files separately.

I will gladly make a PR that adds this feature.

bouk:
I made this PR backwards-compatible, so it still uses the old ways of reading files and directories pre-1.16.

I've made this PR a draft until go 1.16 is released

Resolves #193

### DIFF
--- a/pkg/dbmate/db_116.go
+++ b/pkg/dbmate/db_116.go
@@ -1,0 +1,59 @@
+// +build go1.16
+
+package dbmate
+
+import (
+	"io/fs"
+	"net/url"
+	"os"
+	"path"
+	"time"
+)
+
+// DB allows dbmate actions to be performed on a specified database
+type DB struct {
+	AutoDumpSchema      bool
+	DatabaseURL         *url.URL
+	MigrationsDir       string
+	MigrationsTableName string
+	SchemaFile          string
+	Verbose             bool
+	WaitBefore          bool
+	WaitInterval        time.Duration
+	WaitTimeout         time.Duration
+	FS                  fs.FS
+}
+
+// New initializes a new dbmate database
+func New(databaseURL *url.URL) *DB {
+	return &DB{
+		AutoDumpSchema:      true,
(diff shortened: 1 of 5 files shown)


pkg/dbmate/db_116.go
```diff
@@ -0,0 +1,59 @@
...
+		FS:                  osFS{},
+	}
+}
+
+// osFS is an fs.FS implementation that just passes on to os.Open
+type osFS struct{}
+
+func (c osFS) Open(name string) (fs.File, error) {
+	return os.Open(name)
+}
```
bouk, line 47: This is needed because of https://github.com/golang/go/issues/44286 (`os.DirFS` won't cut it)

pkg/dbmate/db_116.go
```diff
@@ -0,0 +1,59 @@
+// +build go1.16
```
amacneil, line 1: I'm not particularly attached to retaining go v1.15 support, since dbmate is primarily a standalone tool we don't provide explicit backwards compatibility promises when used as a library. Anyone stuck on earlier versions of Go can continue using the current version of dbmate until they are ready to upgrade.

amacneil:
FYI master branch is set up to use 1.16 now

amacneil:
@bouk would love to merge this if you want to make some final updates to the PR.

I suggest removing support for older versions of Go to simplify the PR and testing. We are using Go 1.16 on main branch now.

bouk:
I'm not sure this is a good until https://github.com/golang/go/issues/44286 is cleared up, because the backwards-compatibility hack I have in here right now could break in a future Go version. So I'm going to close this one for now.

Issue #199, opened 2021-02-26
 Database URL does not support plus sign in password


matt-snider:
When a mysql password contains `+`, it appears to not be handled correctly.
 
I am encoding the password that I pass via `--url` so that `+` becomes `%2B`. I can see that after [`url.Parse()` in main.go][main.go] this gets converted back to `+`. The problem is that the mysql `Driver` calls [`url.QueryUnescape()`][mysql.go], after which the `+` becomes a single space.

If I don't encode the `+`, `url.Parse()` leaves it as is, but `url.QueryUnescape()` converts it to a single space, as in the first case.

I've experimented with `url.PathUnescape()`, and when this is used, everything seems to work properly. It seems like this is a safe switch to make since the [documentation][pathunescape] says:

> PathUnescape is identical to QueryUnescape except that it does not unescape '+' to ' ' (space).

I can see that this change was introduced in #76. @kriive care to weigh on this?

[main.go]: https://github.com/amacneil/dbmate/blob/0c758a8d9fdad0359bf352bd18b1d3ec6572dd1e/main.go#L257
[mysql.go]: https://github.com/amacneil/dbmate/blob/0c758a8d9fdad0359bf352bd18b1d3ec6572dd1e/pkg/driver/mysql/mysql.go#L55
[pathunescape]: https://golang.org/pkg/net/url/#PathUnescape

kriive:
As far as my opinion counts, I'm all in (I only contributes that fix to this project). It certainly does not make any sense to parse "+" chars to spaces. Make sure to add a test to cover this case!

amacneil:
Definitely looks like a bug - from memory, the mysql driver expects a plain password string (no percent encoding) which is why we unencode it, but converting `+` to space is obviously not correct.

I added a comment in #200. Thanks!

PR #200, opened 2021-02-26
mysql: Use url.PathUnescape instead of url.QueryEscape so that '+' in passwords works

matt-snider:
This addresses #199 which I just opened. As the [documentation][pathunescape] says:

> PathUnescape is identical to QueryUnescape except that it does not unescape '+' to ' ' (space).

I think that makes sense, since `+` in passwords gets converted to a space otherwise, and this isn't correct.

[pathunescape]: https://golang.org/pkg/net/url/#PathUnescape

### DIFF
--- a/pkg/driver/mysql/mysql.go
+++ b/pkg/driver/mysql/mysql.go
@@ -52,7 +52,7 @@ func connectionString(u *url.URL) string {
 
 	// Get decoded user:pass
 	userPassEncoded := u.User.String()
-	userPass, _ := url.QueryUnescape(userPassEncoded)
+	userPass, _ := url.PathUnescape(userPassEncoded)
 
 	// Build DSN w/ user:pass percent-decoded
 	normalizedString := ""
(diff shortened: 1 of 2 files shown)


amacneil:
Thanks for the PR and bug report! I think this makes sense - it seems like an oversight (I don't think I was even aware of those two separate methods).

There are a bunch of tests for `connectionString()`, can you add another one (or maybe modify the `"special chars"` test) with a comment explaining the correct `+` handling we are looking for?

matt-snider:
@amacneil I added a test, but I sadly couldn't get `make test` to work properly a reasonable amount of time, gave up, and pushed, hoping the CI would tell me if it was correct. I have only written a small amount of Go so the problem is definitely on my end. The CI still hasn't responded, but I guess we'll see shortly whether it was correct!

amacneil:
Strange, I'm not sure why CI didn't trigger. I'll take a look later today!

PR #204, opened 2021-03-13
postgre: add more error information to an error message (position, line number, partial SQL) 

### LINKED ISSUE #102: Lack of verbosity on errors during migrate

silbinarywolf:
**Historical context:**

- Previous Draft: https://github.com/amacneil/dbmate/pull/203
- Issue where I discusses the plan for this PR here: https://github.com/amacneil/dbmate/issues/102#issuecomment-792228369

**Note from exploring this change:**

- MySQL driver unfortunately doesn't give as much information, even if we were to parse the error string, we could only get the line number

Fixes #102


silbinarywolf:
@amacneil Is this okay to merge in?

If not, I don't have a personal stake in this being merged in so I'm fine with this being closed and unmerged. (I feel like the code solution could be less complicated / better but I have no idea how nor do I have the time to explore it)

pkg/dbutil/dbutil.go
```diff
@@ -5,13 +5,61 @@ import (
...
+// This was initially created to work around deficiency in the PostgreSQL drivers for Go where
+// this kind of information must be manually computed
+type DetailedSQLError struct {
+	SQLError error
+	Line     int
+	Column   int
+	Position int
+}
+
+var _ error = new(DetailedSQLError)
```
amacneil, line 27: what is this line for?

silbinarywolf: explicitly asserts at compile-time that DetailedSQLError satisfies the `error` interface.
https://stackoverflow.com/a/34619122/5013410

dqii:
Just wondering, was there a reason this didn't get merged?

atombender:
This looks good to me. If you want to tweak things, @amacneil, it could be done as a follow-up. It doesn't have to perfect.

dossy (changes requested):

dossy:
Given that this issue (originally #102) has recently celebrated its fourth birthday 🎂, I took a stab at reviewing this PR, and left review comments above.

Rather than burden the OP with bringing their branch up to date with all the changes since the PR was created, as well as implementing any requested changes from my review, I did the work and submitted PR #495.

I'd like to propose closing PR #204 in favor of merging PR #495, and thank @silbinarywolf for their initial effort which I borrowed from conceptually.

How do people feel about my proposal?

PR #205, opened 2021-03-16
Update features for golang-migrate

breml:
* Timestamp versioning is possible, create migration files with `-format` flag for `create` command: https://github.com/golang-migrate/migrate/tree/master/cmd/migrate#usage
* There is no separate configuration file needed or even supported: https://github.com/golang-migrate/migrate#cli-usage
* Binary releases are provided from the release page: https://github.com/golang-migrate/migrate/releases


amacneil:
Thanks!

I guess "no separate configuration file" is debatable, since you have to pass everything on the command line it would require you to write a bash script or something remembering your settings, which then becomes a config file separate from the app. One of my design goals for dbmate was to automatically load `.env` files, which are a common place to store app connection strings.

amacneil:
Good to see they have binary releases now. I originally created this comparison table some years ago.

Issue #218, opened 2021-07-08
Having multiple statements in migration throws error on ClickHouse


Rio6:
I have a simple migration to test on ClickHouse like this
```sql
-- migrate:up
select 1;
select 2;


-- migrate:down

```
Running the above migration throws this error
```
Applying: 20210708033543_test.sql
Error: code: 62, message: Syntax error (Multi-statements are not allowed): failed at position 23 (end of query) (line 2, col 9): ;
select 2;


.
```
I'm guess either there's an option needed to set when connecting to the DB, or maybe ClickHouse doesn't support multi-statement query. Perhaps dbmate can split the query up and send them one by one?

ybogo:
I am also stacked with this problem. It seems that clickhouse http client has multiquery support https://clickhouse.tech/docs/en/interfaces/cli/#command-line-options but CH go client still doesn't support multiqueries. My issue in clickhouse go client repo is opened without any answers for several month https://github.com/ClickHouse/clickhouse-go/issues/351
Maybe it is possible to add implicit split in dbmate?

amacneil:
It sounds like the easiest way to solve this is probably to fix it in the ClickHouse go client, since our other drivers support multi statements natively.

It would be possible (but not trivial) to add a SQL parser and send statements one at a time. I'm open to a PR for this.

Rio6:
Yeah it seems like clickhouse-go is the better place for this as it requires parsing the SQL.

ybogo:
> Yeah it seems like clickhouse-go is the better place for this as it requires parsing the SQL.

Unfortunately there will be no multiquery support on the clickhouse-go driver side
https://github.com/ClickHouse/clickhouse-go/issues/351#issuecomment-1013864963

amacneil:
In that case, it might be a showstopper. I'm not very excited about adding SQL parsing logic to dbmate, it would be a lot of work to maintain over time, with inconsistencies between database plugins, etc. So unless there is upstream support for multiquery in the sql driver, we probably can't support it in dbmate.

AntonFriberg:
I have not written a lot of golan but I like the approach taken by golang-migrate https://github.com/golang-migrate/migrate/blob/master/database/clickhouse/clickhouse.go#L139

It simply splits the statements on the `;` character. 

This is not as robust as adding proper SQL parsing logic but it mentions that in the notes. https://github.com/golang-migrate/migrate/tree/master/database/clickhouse#notes

[...]

Issue #74, opened 2019-05-21
Spanner support


amacneil:

AJCStriker:
Hey @amacneil I would be interested in picking this issue up and submitting a PR.

I took a look at the current drivers but is there anything to keep in mind whilst designing an integration? 
In particular clarification about how you view Connection lifecycles would be really helpful?

amacneil:
Last time I looked into this, I ran into the issue that dbmate has been designed around an assumption that all drivers will use `database/sql`, and the spanner driver did not use that. So it would require refactoring the driver interface to remove dependence on `database/sql`.

Long term I think this is a good direction to take dbmate, but just be aware that to achieve this feature I think you will need to refactor the interface and all other drivers to match the new interface, it won't be a simple case of adding one new driver.

From memory there also might be a problem with the spanner driver not accepting multiple sql statements in one string, so we will need to import a sql parsing library and do our own statement splitting. Again I think this is a nice feature that would benefit all drivers eventually (allowing us to give more helpful error messages, etc), but it adds complexity.

PR #219, opened 2021-07-19
add separate migration option

ybogo:
Adds separate migration option.
If true queries will be applied separately one by one. It is useful for databases which does not support multiple queries such as ClickHouse.

Default values for this option is false.

### DIFF
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/amacneil/dbmate/pkg/dbutil"
@@ -314,6 +315,19 @@ func (db *DB) Migrate() error {
 	return db.migrate(drv)
 }
 
+func splitContentOnQueries(content string) []string {
+	var queries []string
+	delimeterSepparator := func(c rune) bool {
+		return c == ';'
+	}
+	queries = strings.FieldsFunc(strings.TrimSpace(content), delimeterSepparator)
+	for i := range queries {
+		queries[i] += ";"
+	}
+
+	return queries
+}
+
 func (db *DB) migrate(drv Driver) error {
 	files, err := findMigrationFiles(db.MigrationsDir, migrationFileRegexp)
 	if err != nil {
@@ -358,11 +372,20 @@ func (db *DB) migrate(drv Driver) error {
 
 		execMigration := func(tx dbutil.Transaction) error {
 			// run actual migration
-			result, err := tx.Exec(up.Contents)
-			if err != nil {
-				return err
-			} else if db.Verbose {
-				db.printVerbose(result)
+			var content []string
+			if !up.Options.separate() {
+				content = append(content, up.Contents)
+			} else {
+				content = splitContentOnQueries(up.Contents)
+			}
+
+			for _, query := range content {
+				result, err := tx.Exec(query)
+				if err != nil {
+					return err
+				} else if db.Verbose {
+					db.printVerbose(result)
+				}
 			}
 
 			// record migration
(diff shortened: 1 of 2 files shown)


ybogo:
@amacneil, is consistent functionality? Take a look pls.
This is related to #218

amacneil:
It looks like you're just splitting statements on semicolons, that is not sufficient. If we add this feature it will need a proper sql parser which can handle semicolon in the middle of comments or strings.

Another edge case to consider is the [MySQL DELIMITER](https://www.mysqltutorial.org/mysql-stored-procedure/mysql-delimiter/) option. I'm not sure whether other databases have similar concepts. Maybe we don't need to support that.

ybogo:
Yes, know I am understand that just splitting on semicolons is not proper way.
I can do splitting using regexps but it looks like that it is very hard to do it using built-in regexps package because it doesn't support lookarounds (to keep O(n) complexity). But this task could be solved with [regexp2](https://github.com/dlclark/regexp2) package. Is it ok to add this third-party dependency?

amacneil:
I don't think a regex will be sufficient either. You need a proper lexer/parser.

There are a few go sql parsers around, e.g. https://github.com/pingcap/parser and https://github.com/blastrain/vitess-sqlparser. I don't know how well they will work across different sql implementations.

I do think this is a valuable feature (several issues have come up over the years caused by not breaking up statements before sending them, and Spanner support (https://github.com/amacneil/dbmate/issues/74) is also blocked on needing sql statement splitting. But if we add it it needs to be implemented in a reliable manner.

ybogo:
> I don't think a regex will be sufficient either. You need a proper lexer/parser.

Why? It seems that it is enough not to split on semicolons which are in quotes or inside a comment. Or I am wrong?

amacneil:
I could be convinced a regex would work with some solid unit tests, but it is a complex enough problem that it should be a separate go package that we import. 

Some edge cases to take into account:
- semicolon inside single or double quote string literals or identifiers
- semicolon inside multi line string literals
- correctly handling nested quotes (e.g. don’t get tripped up by a single quote inside a double quoted string literal)
- semicolon inside comments
- semicolon inside multi line comments 
- don’t get tripped up by quote characters inside comments
- various idiosyncrasies between sql dialects 
- MySQL `DELIMITER` statement which changes the delimiter for future statements (plus not sure if other dialects have similar concepts)
- it seems like at least some sql engines support implicit statement splitting? e.g. no semicolon at all. Not sure whether we need to support that though, maybe it is an outdated usage. 

So - parsing source code (like sql) is a fairly well understood category of problem, so it seems more reliable to use the right tool for the job (lexer) rather than a wildly complicated set of regexes to handle all the above? Either way it’s not something I want to maintain inside the dbmate repo if we can avoid it, I would rather use a well tested third party package.

ybogo:
I think that is not possible to implement implicit statement splitting smoothly neither using regexes nor sql parser. But I found this solution in other project https://github.com/delium/clickhouse-migrator#multi-statement-and-single-statement-migrations
JSON array parsing can be easily implemented and it doesn't look very scary. @amacneil, what do think about this?

amacneil:
I don't particularly want to introduce changes to the migration file format. It might be possible to achieve something similar to your JSON array example by allowing multiple `-- migrate:up` blocks in the same migration file (or some other comment that indicates a separately executed block from the same migration). That does open up questions about how we treat transactions (do we use one transaction per file or one transaction per migration block? what do we do if there is `transaction:true` in one block and `transaction:false` in the other block). If you or someone else wants to experiment with that idea I'm open to considering it (it would need good unit tests and to not break any existing functionality).

Dbmate is basically feature complete at this stage, so we can accept bug fixes and minor changes but I want to avoid making any significant changes. I'm going to close this PR since since the semicolon splitting approach is not going to work.

PR #226, opened 2021-07-26
golang-migrate supports connection timeout

aastein:
golang-migrate has the option `-lock-timeout N` which I think covers `Ability to wait for database to become ready`

https://github.com/golang-migrate/migrate/tree/master/cmd/migrate


amacneil:
Does it work if the database container has not yet been started (e.g. port is closed)? It sounds like this feature is talking about more of a muted, but maybe it will work even if the db is not yet ready.

aastein:
You are correct

PR #230, opened 2021-08-04
Postgres defaults to unix socket with no host

### LINKED ISSUE #229: postgres automatic peer auth
Currently, it's possible to connect to a database on localhost using a domain socket. For example:

```
$ dbmate -u "postgres:///database_name?socket=/var/run/postgresql" status
```

It'd be really nice if this was autodetected and you could just do:

```
$ dbmate -u postgres:///database_name status
```

The following works well on my machine:

```
$ psql postgres:///database_name
```

I believe that `psql` is using the `peer` authentication method on my local machine over the domain socket that postgres exposes.

Some more info:
https://www.postgresql.org/docs/current/auth-peer.html

wraithm:
Closes #229

When there's no host given in a postgres connection url, use the unix
domain default socket. Previously, the default was a localhost TCP
connection. This default behavior is more in line with standard postgres
clients out there, like psql.

### DIFF
--- a/pkg/driver/postgres/postgres.go
+++ b/pkg/driver/postgres/postgres.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"runtime"
 	"strings"
 
 	"github.com/amacneil/dbmate/pkg/dbmate"
@@ -48,7 +49,14 @@ func connectionString(u *url.URL) string {
 
 	// default hostname
 	if hostname == "" {
-		query.Set("host", "/var/run/postgresql")
+		switch runtime.GOOS {
+		case "windows":
+			hostname = "localhost"
+		case "linux":
+			query.Set("host", "/var/run/postgresql")
+		default:
+			query.Set("host", "/tmp")
+		}
 	}
 
 	// host param overrides url hostname
--- a/pkg/driver/postgres/postgres.go
+++ b/pkg/driver/postgres/postgres.go
@@ -50,12 +50,12 @@ func connectionString(u *url.URL) string {
 	// default hostname
 	if hostname == "" {
 		switch runtime.GOOS {
-		case "windows":
-			hostname = "localhost"
 		case "linux":
 			query.Set("host", "/var/run/postgresql")
-		default:
+		case "darwin", "freebsd", "dragonfly", "openbsd", "netbsd":
 			query.Set("host", "/tmp")
+		default:
+			hostname = "localhost"
 		}
 	}
 
--- a/pkg/driver/postgres/postgres.go
+++ b/pkg/driver/postgres/postgres.go
@@ -48,7 +48,7 @@ func connectionString(u *url.URL) string {
 
 	// default hostname
 	if hostname == "" {
-		hostname = "localhost"
+		query.Set("host", "/var/run/postgresql")
 	}
 
 	// host param overrides url hostname
--- a/pkg/driver/postgres/postgres_test.go
+++ b/pkg/driver/postgres/postgres_test.go
@@ -56,7 +56,7 @@ func TestConnectionString(t *testing.T) {
 		expected string
 	}{
 		// defaults
-		{"postgres:///foo", "postgres://localhost:5432/foo"},
+		{"postgres:///foo", "postgres://:5432/foo?host=%2Fvar%2Frun%2Fpostgresql"},
 		// support custom url params
 		{"postgres://bob:secret@myhost:1234/foo?bar=baz", "postgres://bob:secret@myhost:1234/foo?bar=baz"},
 		// support `host` and `port` via url params
@@ -85,11 +85,11 @@ func TestConnectionArgsForDump(t *testing.T) {
 		expected []string
 	}{
 		// defaults
-		{"postgres:///foo", []string{"postgres://localhost:5432/foo"}},
+		{"postgres:///foo", []string{"postgres://:5432/foo?host=%2Fvar%2Frun%2Fpostgresql"}},
 		// support single schema
-		{"postgres:///foo?search_path=foo", []string{"--schema", "foo", "postgres://localhost:5432/foo"}},
+		{"postgres:///foo?search_path=foo", []string{"--schema", "foo", "postgres://:5432/foo?host=%2Fvar%2Frun%2Fpostgresql"}},
 		// support multiple schemas
-		{"postgres:///foo?search_path=foo,public", []string{"--schema", "foo", "--schema", "public", "postgres://localhost:5432/foo"}},
+		{"postgres:///foo?search_path=foo,public", []string{"--schema", "foo", "--schema", "public", "postgres://:5432/foo?host=%2Fvar%2Frun%2Fpostgresql"}},
 	}
 
 	for _, c := range cases {
--- a/pkg/driver/postgres/postgres_test.go
+++ b/pkg/driver/postgres/postgres_test.go
@@ -53,12 +53,12 @@ func TestGetDriver(t *testing.T) {
 
 func defaultConnString() string {
 	switch runtime.GOOS {
-	case "windows":
-		return "postgres://localhost:5432/foo"
 	case "linux":
 		return "postgres://:5432/foo?host=%2Fvar%2Frun%2Fpostgresql"
-	default:
+	case "darwin", "freebsd", "dragonfly", "openbsd", "netbsd":
 		return "postgres://:5432/foo?host=%2Ftmp"
+	default:
+		return "postgres://localhost:5432/foo"
 	}
 }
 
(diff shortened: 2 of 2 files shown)


wraithm:
This works on my machine, btw

wraithm:
@amacneil, the only thing that I think one might object to with this PR is Windows support. My understanding is that Windows 10 Postgres does support UNIX domain sockets, but there's no default path like there is on all the unix-likes. You can always say `postgres://localhost/` to get a TCP/IP connection. `postgres:///` will likely just fail on Windows. I'm not super familiar with golang, but we could put a thing that defaults to localhost if we're on a Windows machine.

Lemme know what you think.

wraithm:
Oh okay, I also just checked postgresql installed via brew on macOS, and it defaults to `/tmp/` for the socket dir. It looks like `/tmp/` might be the general default, but then Ubuntu, which is what I'm on, your docker image and many other linux distros change the default socket location to `/var/run/postgresql`.

wraithm:
Okay, so, the plot thickens:

https://git.launchpad.net/ubuntu/+source/postgresql-12/tree/debian/patches/51-default-sockets-in-var.patch

ubuntu/debian patch `libpq` such that the default folder is `/var/run/postgresql`. The default from the postgres sources themselves is `/tmp`. It looks like other distros do similar things. `libpq` will just magically work on whatever platform you're on, it seems.

To mimic the behavior, we could just try all of the above until something works?

amacneil:
Hmm, that's not ideal. Dbmate has specific options around failed connections and waiting for a connection to become available, so I don't think attempting multiple different connections is feasible.

One option might be to just assume `/var/run/postgresql` on linux, and `/tmp` on mac/windows? Do you happen to know what other common linux distros use?

wraithm:
Yeah, I think that makes a ton of sense. That's a very easy fix. Let's go with `/var/run/postgresql` on Linux and `/tmp` otherwise.

Here's a little table of distros and where they put these sockets:

|Distro|DEFAULT_PGSOCKET_DIR|source|
|-----|------|------|
|Ubuntu/Debian|/var/run/postgresql|https://git.launchpad.net/ubuntu/+source/postgresql-12/tree/debian/patches/51-default-sockets-in-var.patch |
|Fedora|/var/run/postgresql|https://src.fedoraproject.org/rpms/postgresql/blob/rawhide/f/postgresql-var-run-socket.patch#_50 |
|Gentoo|/run/postgresql|https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-db/postgresql/postgresql-12.7.ebuild#n95 |
|Archlinux|/run/postgresql|https://github.com/archlinux/svntogit-packages/blob/packages/postgresql/trunk/postgresql-run-socket.patch#L9 |
|OpenSUSE|/var/run/postgresql|https://build.opensuse.org/package/view_file/openSUSE:Factory/postgresql13/postgresql-var-run-socket.patch?expand=1 |

It looks like the vast majority of Linux distros (Debian/RedHat derived) use `/var/run/postgresql`. Arch and Gentoo use `/run/postgresql`. I think this is acceptable. I'll fix up my PR to do that.

wraithm:
@amacneil, updated. Thanks for your input. This default behavior wouldn't be a problem if dbmate used a postgresql driver based on linking against libpq, the official postgresql client library. I'm not that familiar with the golang ecosystem, but it seems strange to me that all of the postgresql drivers I can find are "pure Go" and don't just wrap libpq. There are two libpq cgo wrapper libs that I can find, but last updated in 2012 and 2014 sadly.

amacneil:
Thanks for the table. This approach looks good.

I think we should make the tests work across platform. As written I think they will fail on non-linux platforms? Maybe it's possible to mock runtime.GOOS, or inject it somehow. I don't write much go these days either. @wkalt how would you make these tests work cross-platform?

Re linking official libpq - I can't speak to why there aren't any well maintained cgo wrapper projects. I guess people just like avoiding cgo where necessary. It wouldn't help in our specific case because we statically link the linux release binary (so that people can run it on alpine linux etc with no dependencies), so I assume we would

wraithm:
Ah, that makes a ton of sense. 👍 Thanks for that explanation!

I confirmed that at least this PR works as expected on both Ubuntu 20.10 and macOS Big Sur.

What platforms were you interested in testing? Making these tests run on platforms that aren't Linux could be pretty tricky, no? Docker is pretty Linux targeted.

I haven't seriously evaluated this. You miiiight be able to do macOS: https://github.com/sickcodes/Docker-OSX

I think testing Windows locally would be very difficult, even with using Vagrant or something like that.

Maybe GitHub actions would do the trick? I've never used Window or macOS in GitHub actions, but I'd imagine that could work. It might be very painful to set that up, slow trial and error, but possible.

amacneil:
Does `make test` work on Mac? I would be surprised since your tests hard code the /var expectation. I can try it later on my mac too. 

I am thinking about running the other platforms in CI on GitHub Actions (#231). But that doesn’t need to block this PR, a manual validation will be fine.

wraithm:
You know what I'll do? I'll make the test dependent on GOOS.

wraithm:
I think this should work. The tests pass in docker. I couldn't get this to work well on my mac because I don't have clickhouse/mysql/etc on my mac. However, I'm pretty confident that these tests should work on both macOS and Windows. Let me know what you think.

wkalt:
@amacneil  @wraithm Couple thoughts on this - 

1. If going with the GOOS switch approach, it seems more conservative to default to the previous behavior host = "localhost" to ensure we're not breaking existing functionality on some of the more obscure OS's. Possible values for GOOS are listed here - you can see there are a good number more to cover than windows/linux/darwin: https://github.com/golang/go/blob/master/src/go/build/syslist.go#L10 .
2. The location of the socket is a configuration option in postgresql.conf called `unix_socket_directories` that the user is free to change. It would be safer to dispatch on this than on GOOS. The change here will break for users that have changed this setting, if they are relying on the current behavior when host is not specified, supposing such users exist. This value can be assessed from the config file (supposing the location is known, which it is not in general) or from the server by executing `show unix_socket_directories` in psql or another client.

wraithm:
@wkalt 

1. You're totally right. I'll look over that syslist and try to fit to what each one does.

I'll probably do:

|OS|host|
|--|----|
|linux|/var/run/postgresql|
|*bsd/macOS|/tmp|
|otherwise (incl. windows)|localhost|

2. `unix_socket_directories` is for the server's config. Most (all?) clients don't look at `unix_socket_directories` and just go with a hardcoded default, including the official `libpq`. If the default unix socket isn't there, then the clients tend to error. That's the expected behavior. The user of the client lib must specify where the socket is if the socket isn't in the standard place.

wraithm:
@wkalt @amacneil

Is there a list of supported platforms for dbmate? I think it'd be great to narrow my scope to a specific set of supported platforms, and then go with `localhost` for everything else, just to be conservative. I see that there are builds for linux, macOS, and Windows. I don't see anything in the code that would really stop you from using any of the BSDs. Are people using postgres/dbmate on AIX or z/OS 😂 ?

wraithm:
@wkalt @amacneil,

I looked at the intersection of postgresql's [supported platforms](https://www.postgresql.org/docs/13/supported-platforms.html) and Go's [syslist](https://github.com/golang/go/blob/master/src/go/build/syslist.go#L10).

postgresql: `Linux (all recent distributions), Windows (XP and later), FreeBSD, OpenBSD, NetBSD, macOS, AIX, HP/UX, and Solaris`
syslist: `aix android darwin dragonfly freebsd hurd illumos ios js linux nacl netbsd openbsd plan9 solaris windows zos`

That leaves us with: `aix darwin dragonfly freebsd linux netbsd openbsd solaris windows` for `GOOS` that we have to handle.  (DragonflyBSD is basically FreeBSD, they use the same packaging system)

I found out that [AIX's unix domain sockets are broken](https://www.postgresql.org/docs/10/installation-platform-notes.html#INSTALLATION-NOTES-AIX), lmao 😂, so `localhost` there. I couldn't make heads or tails of what SmartOS/Illumos/Solaris do. `/tmp` would likely work, but I went with `localhost` to be conservative. BSDs are all `/tmp`. Most linux distros do `/var/run/localhost`.

Thanks @wkalt for pointing out that we should be more careful here.

wraithm:
@amacneil @wkalt, any thoughts here?

wraithm:
@amacneil ping. What do you think? This has been working well for us.

wraithm:
We kinda ran into another case today where this would be really useful for us to have on linux. My engineers keep getting permission errors when using postgresql urls that work with `psql` or `pgcli`, but don't work with `dbmate`. I've been running this version locally and it works great for me on Ubuntu, but it'd be great to not have to run a fork of dbmate for the rest of the team.

Thank you for your amazing work on dbmate! dbmate has been a great tool for us so far.

wraithm:
@amacneil @wkalt, Sorry to be annoying, but is there anything I can do to push this along?

amacneil:
Hi, thanks for the ping! Sorry, I haven't had much time for dbmate maintenance lately.

Merging now :) Thanks for the detailed research into postgres defaults across various platforms.

wraithm:
Awesome! Thank you ☺️

PR #241, opened 2021-10-02
Add multiarch docker images

Aposhian:
This publishes multiarch (arm64 and amd64) images to DockerHub and GitHub Container Registry


amacneil:
Thank you! I recently landed https://github.com/amacneil/dbmate/pull/231 which vastly simplifies our build process (no more cross-OS cross compiling), and I've added some commits to your PR to merge these changes.

.github/workflows/ci.yml
```diff
@@ -75,16 +75,24 @@ jobs:
...
           set -x
           docker version
+          docker buildx version
           docker-compose version
 
       - name: Build docker image
         run: |
           set -x
-          docker-compose build
+          docker buildx bake --load --set "*.platform=linux/amd64"
```
amacneil, line 95: What does this do / why is it necessary?

.github/workflows/ci.yml
```diff
@@ -103,14 +111,43 @@ jobs:
...
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          target: release
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v') }}
+          tags: ${{ steps.meta.outputs.tags }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
amacneil, line 153: Would this cache be used for the previous docker build step? If not, it seems like it would do more harm than good (building twice with different caches rather than reusing the same cache).

I'm going to remove it for now until I can do further testing, but let me know your thoughts.

PR #242, opened 2021-10-06
Make Status read only

Jell:
So we have this use case where our developers have read-only access to our databases, and want to be able to run `dbmate status` before running a migration as part of our CI/CD pipeline.

Surprisingly, `dbmate status` has the potential side-effect of creating the migrations table if it's not there, which fails for a read only user.

This change makes the Status command read only by checking if the table exists, and returning all migrations as pending if it doesn't (and works as expected otherwise).

### DIFF
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -576,15 +576,24 @@ func (db *DB) CheckMigrationsStatus(drv Driver) ([]StatusResult, error) {
 		return nil, fmt.Errorf("no migration files found")
 	}
 
-	sqlDB, err := db.openDatabaseForMigration(drv)
+	sqlDB, err := drv.Open()
 	if err != nil {
 		return nil, err
 	}
 	defer dbutil.MustClose(sqlDB)
 
-	applied, err := drv.SelectMigrations(sqlDB, -1)
+	applied := map[string]bool{}
+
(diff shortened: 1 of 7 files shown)


Jell:
anything I could do to help move this forward?

Enrico204:
Interesting. I agree, status should be R/O, there is no need for it to create a table

pkg/driver/sqlite/sqlite_test.go
```diff
@@ -233,6 +233,7 @@ func TestSQLiteSelectMigrations(t *testing.T) {
 	// test limit param
 	migrations, err = drv.SelectMigrations(db, 1)
 	require.NoError(t, err)
+
```
Enrico204, line 236: You may want to remove this space, to avoid creating conflicts where there is no need to :-)

Jell: fixed!

amacneil:
Would be good to add a test making sure the table doesn’t get created before we merge

Jell:
> Would be good to add a test making sure the table doesn’t get created before we merge

Done! https://github.com/amacneil/dbmate/pull/242/commits/7c76b9b3b15ef1ea30c0dd3c90ffb6b9745003ca

amacneil:
Thanks!

PR #246, opened 2021-11-13
Add Migration Summary

tmelliottjr:
- Adds output for total skipped and applied migrations

### DIFF
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -341,10 +341,13 @@ func (db *DB) migrate(drv Driver) error {
 		return err
 	}
 
+	var totalSkipped, totalApplied int
+
 	for _, filename := range files {
 		ver := migrationVersion(filename)
 		if ok := applied[ver]; ok {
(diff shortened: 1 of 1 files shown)


amacneil:
Have you found this necessary? I don't particularly like the idea of listing how many skipped migrations (it makes it sound like there was a problem, which is not the case), and we already show you the applied migrations.

amacneil:
I'm going to close this PR since I don't personally see the value in this. If others feel differently, please chime in with use cases.

tmelliottjr:
> Have you found this necessary? I don't particularly like the idea of listing how many skipped migrations (it makes it sound like there was a problem, which is not the case), and we already show you the applied migrations.

I wouldn't call it necessary, but my team saw value in this when running as part of our pipelines. At current if the tool runs successfully and no migrations were applied, there isn't any output. The indication that the tool ran successfully and skipped/applied migrations was helpful for getting an idea of what was happening at-a-glance.

Issue #252, opened 2021-12-31
Proposal: Having separation in timestamp formatting for readability


gmhafiz:
## Issue 

Currently dbmate creates a new migration file with the following [format](https://github.com/amacneil/dbmate/blob/main/pkg/dbmate/db.go#L245)

```go
timestamp := time.Now().UTC().Format("20060102150405")
```

When you have a bunch of them it's hard to easily find a migration made in a specific month.

I propose having some separation between the year, month and day in the timestamp. For example:

 - `2006_01_02_150405`
 - `2006-01-02-150405`

These are still sortable (if all follows the same format). Complies with [`migrationVersion()`](https://github.com/amacneil/dbmate/blob/main/pkg/dbmate/db.go#L448) regex.

In my opinion, it is much easier to read the dates since we care YYYY_MM_DD the most.

![image](https://user-images.githubusercontent.com/1927253/147804563-536744f4-6bf5-4ac4-8faf-ef36cbd75527.png)

## Problem 

Obviously it it not backward-compatible because sorting with both separated and non-separated timestamp will not sort by date.

Unless something is done to parse the timestamp in place of [`sort.Strings(matches)`](https://github.com/amacneil/dbmate/blob/main/pkg/dbmate/db.go#L425) and sort them by date accordingly (which is easy), then, it can be backward compatible. I am unsure the impact to the records saved in `schema_migrations` table though.

gmhafiz:
It appears [migrationversion](https://github.com/amacneil/dbmate/blob/main/pkg/dbmate/db.go#L448) is also affected. I handled that too in the following pull request.

https://github.com/amacneil/dbmate/pull/253

gmhafiz:
I beginning to think that simply using `time.RFC3339` would be better.

- no spaces
- part of rfc standard and standard library
- includes timezone
- still can differentiate the year, month, day
- time is separated by colons

Once issue is how colon is handled under Windows because I remember powershell had difficulties a string containing colons needing to escape with backslash.

amacneil:
The migration version isn't really supposed to be a human readable date, since the date is irrelevant to the migration contents. We just use a date-based version number so that two developers on different branches don't try to create "migration version 3".

[...]

PR #253, opened 2021-12-31
Add ability to use timestamp format with underscore from cli argument

gmhafiz:
Based on https://github.com/amacneil/dbmate/issues/252.

Reading timestamp that contains a separator is easier to read and find out the year, month, and day. For example:

 20211231054630_create_users_table.sql
vs
 2021_12_31_054630_create_users_table.sql


gmhafiz:
It defaults to `20211231054630_create_users_table.sql` format.

Allows the option of user to use the new timestamp format by supplying cli argument or environment variable.

It is possible to mix between old and new timestamp format since the `sort()` function handles sorting only on the timestamp. Validation done through unit testing

amacneil:
Thanks for the PR, and I hope it meets your needs. However, as discussed in https://github.com/amacneil/dbmate/issues/252#issuecomment-1015086331 I don't think this is a good idea, because it makes it much harder to differentiate between the migration version number and the migration name.

I don't see a good path forward here - if you need this it's probably better to use a modified version of dbmate. We could potentially support period separators in the migration version number, but that also adds complexity to dbmate and goes against our philosophy of keeping things easy and simple.

PR #266, opened 2022-01-27
Fix whitespace issues in SQLite absolute paths in DATABASE_URL

### LINKED ISSUE #264: DATABASE_URL doesn't support whitespaces
I was testing the tool with a path containing a whitespace and this happened

```
Error: unable to open database file: no such file or directory
```

## Reproduce

`dbmate-x86_64-apple-darwin -u "sqlite:/Users/USERNAME/Library/Application Support/store.sqlite3" status`

Enrico204:
When the DSN is in the form `sqlite:/absolute/path` or `sqlite://absolute/path` or `sqlite:///absolute/path`, url.Parse
will consider the file path as:
 - `absolute` as the hostname
 - `path` (and the rest until `?`) as the URL path.
Instead, when the DSN is in the form "sqlite:", the (relative) file path is stored in the "Opaque" field.
See: https://pkg.go.dev/net/url#URL

While Opaque is not escaped, the URL Path is. So, in absolute paths the full file path is being escaped as if it was an URL path.
This commit works around this behavior in `ConnectionString()` (`sqlite.go`) so that the return string is always a correct (url-unescaped) file path.

Fixes #264 .

### DIFF
--- a/pkg/dbutil/dbutil.go
+++ b/pkg/dbutil/dbutil.go
@@ -151,3 +151,18 @@ func MustParseURL(s string) *url.URL {
 
 	return u
 }
+
+// MustUnescapePath unescapes a URL path, and panics if it fails.
+// It is used during in cases where we are parsing a generated path.
+func MustUnescapePath(s string) string {
+	if s == "" {
+		panic("missing path")
+	}
+
+	path, err := url.PathUnescape(s)
+	if err != nil {
+		panic(err)
+	}
+
+	return path
+}
--- a/pkg/driver/sqlite/sqlite.go
+++ b/pkg/driver/sqlite/sqlite.go
@@ -47,6 +47,23 @@ func ConnectionString(u *url.URL) string {
 	newURL := *u
 	newURL.Scheme = ""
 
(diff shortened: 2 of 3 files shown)


Enrico204:
I fixed the error in the pipeline by editing an assert for an error. The nature of the error was not altered - the change was actually due to the fact that the error have the path embedded in the string, and now the (invalid) path changed a bit.

(the goal of that test was to assert the error opening a invalid/non existent file, regardless of which file it was. So, there should be no issue with that)

The path change actually comes from my edits in `ConnectionString`. I decided to avoid fixing `ConnectionString` because:
1. the path `.` is invalid for SQLite [1]
2. other cases (relative paths, absolute paths) are handled correctly, and tested

[1] the current behavior for `.` path is now: `ConnectionString(".") == "/."`. The previous one was `ConnectionString(".") == "."`

amacneil:
Spaces in a URL aren’t actually valid (should use `%20`). But this makes sense if people want to copy paste a local path.

amacneil:
Do we have any tests for the `sqlite://foo/bar.sqlite` (two leading slashes) behavior you described? I can’t see any.

PR #267, opened 2022-01-27
Fix UNIX socket support for MySQL in dump action

### LINKED ISSUE #235: dump fail without setting hostname and passoword in MySQl
I just test dump function with mysql instance, that will fail with assign unix socket path.
And the code as below, there are not any args about socket ,only hostname and port.
The general command of mysqldump is  `mysqldump -uroot -S /data/mysql/run/mysql.sock`
```
func (drv *Driver) mysqldumpArgs() []string {
	// generate CLI arguments
	args := []string{"--opt", "--routines", "--no-data",
		"--skip-dump-date", "--skip-add-drop-table"}

	if hostname := drv.databaseURL.Hostname(); hostname != "" {
		args = append(args, "--host="+hostname)
	}
	if port := drv.databaseURL.Port(); port != "" {
		args = append(args, "--port="+port)
	}
	if username := drv.databaseURL.User.Username(); username != "" {
		args = append(args, "--user="+username)
	}
	// mysql recommends against using environment variables to supply password
	// https://dev.mysql.com/doc/refman/5.7/en/password-security-user.html
	if password, set := drv.databaseURL.User.Password(); set {
		args = append(args, "--password="+password)
	}

	// add database name
	args = append(args, dbutil.DatabaseName(drv.databaseURL))

	return args
}
```

Enrico204:
`mysqldumpArgs` was missing the `--socket` parameter when launching `mysqldump` for UNIX socket connections. This pull request adds the parameter if the `socket` query option is added in the database URL

Closes #235


amacneil:
Thanks! Can you add "Closes #235" in the issue description so that it is linked to the bug and will automatically close it when merged?

pkg/driver/mysql/mysql.go
```diff
@@ -145,6 +145,9 @@ func (drv *Driver) mysqldumpArgs() []string {
 	if password, set := drv.databaseURL.User.Password(); set {
 		args = append(args, "--password="+password)
 	}
+	if socket := drv.databaseURL.Query().Get("socket"); drv.databaseURL.Hostname() == "" && socket != "" {
```
amacneil, line 148: I forget, do we have any tests for this function? Would be nice to add a test to cover the new functionality.

Enrico204: Currently, no. But I might try to come up with some testing. Let me see what I can do :-)

amacneil:
Thanks! I added some tests.

Issue #128, opened 2020-04-17
Invoke dbmate from Docker and trigger migration in postgres 


jayarjo:
I was trying the following:
```
$ docker run --rm -it --env DATABASE_URL="..." --mount type=bind,source=...,target=/app amacneil/dbmate --no-dump-schema --migrations-dir /app up
```

However I repeatedly get:

```
Error: dial tcp 127.0.0.1:57806: connect: connection refused
```

`DATABASE_URL` points to the postgres server running on another Docker container (not sure how that should matter?)

Do I need to to anything in addition to the above command to be able to run migrations on another Docker container that is running postgres?

jayarjo:
In fact forget about another Docker container, it fails to connect to the postgres on the host as well, with the same error.

amacneil:
By default docker containers have their own network and IP address. There are several ways you can work around this.

If you're running postgres on your local computer you should be able to connect to it by running the dbmate container with [host networking](https://docs.docker.com/network/host/) (I'll add this to the readme):

```sh
docker run --rm --network=host -it amacneil/dbmate
```

If you're running postgres in another container, again there are several ways to solve this. One option is to expose the port from the postgres container to your host (using `-p` flag), then use host networking (as above) when running dbmate:

```sh
docker run --rm -it -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres
docker run --rm -it --network=host -e DATABASE_URL="postgres://postgres:postgres@pg/foo?sslmode=disable" amacneil/dbmate create
```

Another is to put both the postgres container and dbmate on a private container network (and not expose them to the host):

```sh
docker network create mynet
docker run --rm -d -e POSTGRES_PASSWORD=postgres --name=pg --network=mynet postgres
docker run --rm --network=mynet -e DATABASE_URL="postgres://postgres:postgres@pg/foo?sslmode=disable" amacneil/dbmate create
```

Finally, you can also do this using docker compose by adding `depends_on`, which is what [we do](https://github.com/amacneil/dbmate/blob/master/docker-compose.yml) for dbmate development & testing.

jayarjo:
@amacneil thanks for taking your time to clarify this. Would have been great to have this somewhere in readme.

amacneil:
Thanks for the feedback - I updated the readme.

PR #268, opened 2022-01-27
Add sqitch comparison

### LINKED ISSUE #249: Comparison of dbmate vs sqitch or flyway
Both [sqitch](http://sqitch.org) and [flyway](https://flywaydb.org/download/community) are mature alternatives that would be good to add to your comparison, although they have somewhat larger requirements (Perl for sqitch, Java for flyway).

halostatue:
Resolves #249 (for sqitch at least)

I recommend going through this commit by commit, as I changed some of the formatting in the README. I can undo those now that my notes have been added, but I think that they make the README easier to _edit_ without making it harder to read (either raw or rendered).

I am impressed with what I see of dbmate, and think it’s better than the _other_ tools that you have listed. I am not as convinced that it’s _better_ than sqitch, but it is _simpler_ than sqitch (there is a fairly large learning curve to sqitch in the same way as there’s a large learning curve to vim or git), which is worthwhile.

I think that it would be worthwhile for you to study sqitch and figure out what you can pinch to make dbmate better without compromising your simplicity. The first suggestion would be an optional verify step, which would be run after the deploy step and would invoke the revert step if it fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,26 +399,35 @@ You can customize the name of this table using the `--migrations-table` flag or 
 
 Why another database schema migration tool? Dbmate was inspired by many other tools, primarily [Active Record Migrations], with the goals of being trivial to configure, and language & framework independent. Here is a comparison between dbmate and other popular migration tools.
 
-|                                                              | [dbmate] | [goose] | [sql-migrate] | [golang-migrate] | [activerecord] | [sequelize] |
-| ------------------------------------------------------------ | :------: | :-----: | :-----------: | :--------------: | :------------: | :---------: |
+|                                                              | [dbmate] | [goose] | [sql-migrate] | [golang-migrate] | [activerecord] | [sequelize] | [sqitch] |
+| ------------------------------------------------------------ | :------: | :-----: | :-----------: | :--------------: | :------------: | :---------: | :------: |
 | **Features**                                                 |
-| Plain SQL migration files                                    |    ✅    |   ✅    |      ✅       |        ✅        |                |             |
-| Support for creating and dropping databases                  |    ✅    |         |               |                  |       ✅       |             |
-| Support for saving schema dump files                         |    ✅    |         |               |                  |       ✅       |             |
-| Timestamp-versioned migration files                          |    ✅    |   ✅    |               |        ✅        |       ✅       |     ✅      |
-| Custom schema migrations table                               |    ✅    |         |      ✅       |                  |                |     ✅      |
-| Ability to wait for database to become ready                 |    ✅    |         |               |                  |                |             |
-| Database connection string loaded from environment variables |    ✅    |         |               |                  |                |             |
-| Automatically load .env file                                 |    ✅    |         |               |                  |                |             |
-| No separate configuration file                               |    ✅    |         |               |        ✅        |       ✅       |     ✅      |
-| Language/framework independent                               |    ✅    |    ¹    |       ¹       |        ✅        |                |             |
+| Plain SQL migration files                                    |    ✅    |   ✅    |      ✅       |        ✅        |                |             |   ✅²    |
+| Support for creating and dropping databases                  |    ✅    |         |               |                  |       ✅       |             |          |
+| Support for saving schema dump files                         |    ✅    |         |               |                  |       ✅       |             |          |
+| Timestamp-versioned migration files                          |    ✅    |   ✅    |               |        ✅        |       ✅       |     ✅      |    ³     |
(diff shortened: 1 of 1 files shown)


Enrico204:
LGTM! For the change in the style of the links, I don't have a strong opinion (yet). Let's see what Adrian thinks about it :-)

Thanks, I didn't know sqitch before :-D

README.md
```diff
@@ -66,15 +66,15 @@ $ sudo chmod +x /usr/local/bin/dbmate
 
 **Docker**
 
-Docker images are published to both Docker Hub ([`amacneil/dbmate`](https://hub.docker.com/r/amacneil/dbmate)) and Github Container Registry ([`ghcr.io/amacneil/dbmate`](https://ghcr.io/amacneil/dbmate)).
+Docker images are published to both Docker Hub ([`amacneil/dbmate`]) and Github Container Registry ([`ghcr.io/amacneil/dbmate`]).
 
-Remember to set `--network=host` or see [this comment](https://github.com/amacneil/dbmate/issues/128#issuecomment-615924611) for more tips on using dbmate with docker networking):
```
amacneil, line 71: I've never seen this trick of moving the links to the bottom of the markdown file, but I'd prefer to keep things the way they are. Otherwise, it's hard to know whether a link is missing, or what text like "this comment" are pointing to. If there is other readme formatting cleanup, let's do that in a separate PR so it's easy to review.

halostatue: Can I suggest using the named link trick for the alternatives table, as it makes the table _smaller_ to edit? With the in-line links and `:white_check_mark:` &c. emoji, the table wraps all the way down to 5pt, at which point the rest of the text is unreadable.

Other options exist for the formatting of the alternatives:

1. Pivot the table so that the _features_ are columns and the alternatives are the rows.
2. Change it from a table to a list.

The first (pivoting) is _probably_ the easier to do and will support adding other types pretty easily. I’m not sure how many columns would be included.

halostatue: Also, with respect to this particular one, it probably should be a better reference than `this comment`, Say `[this comment][128-comment615924611]`, which would then be `[128-comment615924611]: (https://github.com/amacneil/dbmate/issues/128#issuecomment-615924611)`.

I use this extensively in most of the files that I maintain for projects, like [mime-types-data/History.md](https://github.com/mime-types/mime-types-data/blob/main/History.md).

amacneil: More generally it's better if we can separate reformatting PRs from PRs that actually change the content. So if we want to explore reformatting the comparison table I think we should do that _after_ a very simple PR to "add sqitch".

halostatue: I think that what I’m saying is that I could not "add sqitch" without reformatting the table at least minimally. It’s too large because of the way that it has been built so far.

amacneil: That's fine - if you use vscode, prettier should reformat it for you. It's easy enough to view the diff without whitespace to see the changes.

I still prefer not moving all the links to the bottom of the file, because it makes it much harder to see if you have created a broken link (e.g. update the text and forget to update the link).

README.md
```diff
@@ -397,30 +397,39 @@ You can customize the name of this table using the `--migrations-table` flag or
...
+>
+> ² Sqitch uses three separate migration in folders: up, down, and verify, which validates that the changes were applied correctly. Transactions are not applied automatically in the files, but the generated templates by default include transactions that can be removed as required.
+>
+> ³ Sqitch migration files are not timestamped, but they are recorded in a plan file and can have prerequisites specified. Sqitch records all migration actions (forward or backward) in multiple tables in the `sqitch` namespace, which can be customized.
+>
+> ⁴ Sqitch is framework independent, but requires a Perl installation or the use of a Docker image to run (which contains the Perl installation).
+>
+> ⁵ Sqitch has support for Firebird, Exasol, Oracle, Snowflake, and Vertica. Adding additional databases is possible but non-trivial. Sqitch also has direct support for integrating pgTAP (PostgreSQL unit tests) and support for Perl's App::Template so that common migrations can be templated (if you have tables that have the same basic definition). There are many other features to Sqitch.
```
amacneil, line 430: I appreciate the detail in these footnotes, but I want to keep this table simple because it can get out date quickly. The primary purpose of this table is to answer "why did you create another database migration tool", not provide a detailed comparison (which would be better in a blog post).

- (2) can be removed, it seems like Sqitch uses sql files to just leave the checkmark
- (3) can be removed, it's too much detail about the functionality of a third party 
- (4) I think it's fine to leave the checkmark and remove the footnote - interpreted languages always need a language interpreter, so it's a bit different to the (1) Go footnote (which it looks like is actually out of date anyway)
- (5) can be removed

Issue #285, opened 2022-03-24
transaction:false not working on PostgreSQL


kashifsoofi:
Hi,
We are using dbmate for migrations, we are running into an issue where we want to run following
```
-- migrate:up transaction:false

CREATE INDEX CONCURRENTLY index_name ON table(new_id);
```

We are getting following error
`Error: pq: CREATE INDEX CONCURRENTLY cannot run inside a transaction block`

I didn't had to chance to run dbmate from source, has anyone else run into similar?

Thanks

rohitpaulk:
@kashifsoofi could you confirm what version of dbmate you're running?

rohitpaulk:
The `transaction:false` option was added in https://github.com/amacneil/dbmate/releases/tag/v1.5.0, I want to make sure that you're running a version that includes that change.

kashifsoofi:
I have tested with `dbmate version 1.11.0` and `dbmate version 1.14.0`, from code it looks like it should not create a transaction

rohitpaulk:
Got it, thanks! Will see if I can replicate.

kashifsoofi:
If I run a single CREATE INDEX CONCURRENTLY statement then there is no error

kashifsoofi:
Modifying test file in dbmate code

```
-- migrate:up
create table users (
  id integer,
  name varchar(255)
);
insert into users (id, name) values (1, 'alice');

CREATE INDEX CONCURRENTLY index_users_on_id ON users(id);

-- migrate:down
drop table users;
```

kashifsoofi:
I think the solution would be to somehow run each statement 1 by 1 incase there are multiple statements in a migration file.

kashifsoofi:
Something like this
```
func (m migrationOptions) MultiStatement() bool {
	if val, ok := m["multi-statement"]; ok {
		return val != "true"
	}

	return false
}
```
And then in db.go
```
...
			err = doTransaction(sqlDB, execMigration)
		} else if up.Options.MultiStatement() {
			var stmtSeparatorRegExp = regexp.MustCompile(`;`)
			stmts := stmtSeparatorRegExp.Split(up.Contents, -1)

			for i, stmt := range stmts {
				err = execMigrationStmt(sqlDB, stmt, i+1 == len(stmts))
				if err != nil {
					break
				}
			}
		} else {
			// run outside of transaction
...
```

rohitpaulk:
[...]

PR #286, opened 2022-03-24
#285 Add support for multi-statement

kashifsoofi:
Multi Statement support for running migration script statement by statement. This is to support running multiple statements e.g. create index concurrently in postgresql in a single migration file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,18 @@ ALTER TYPE colors ADD VALUE 'orange' AFTER 'red';
 
 `transaction` will default to `true` if your database supports it.
 
+**multi-statement**
+
+`multi-statement` is useful if you need to run some SQL which cannot be executed from within a block. For example, in Postgres, you would need to run create index concurrently statements on its own, same can be achieved by having a single create index concurrently statement by setting transaction off:
+
+```sql
+-- migrate:up multi-statement:true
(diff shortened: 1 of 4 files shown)


amacneil:
We have discussed splitting statements before in https://github.com/amacneil/dbmate/pull/219#issuecomment-890414357 and I do not think it is a good idea to support. Parsing SQL is much more complicated than simply matching `;` in a regex.

I don't quite understand why you are seeing issues in https://github.com/amacneil/dbmate/issues/285 but it sounds like a bug in `transaction:false` rather than a completely new setting we need?

kashifsoofi:
> We have discussed splitting statements before in [#219 (comment)](https://github.com/amacneil/dbmate/pull/219#issuecomment-890414357) and I do not think it is a good idea to support. Parsing SQL is much more complicated than simply matching `;` in a regex.
> 
> I don't quite understand why you are seeing issues in #285 but it sounds like a bug in `transaction:false` rather than a completely new setting we need?

Agree Its not simple as matching ';' in a regex.

I think their is no bug in `transaction:false` but rather how pg driver/postgres runs multiple statements in `Exec` I think it automatically encloses it in a transaction.

Anyway happy to close this PR.

PR #291, opened 2022-04-16
Do not discard host parameter in postgresql URLs

chloekek:
It currently seems to be impossible to specify a socket path other than the default.


chloekek:
Nevermind, this fix is inadequate.

amacneil:
Did you try this syntax from the readme?

```
DATABASE_URL="postgres://username:password@/database_name?socket=/var/run/postgresql"
```

Issue #292, opened 2022-04-26
rollback: export ErrNoRollback to rollback all


gnuletik:
Hi,

First, thanks for this great project!

I'm integrating dbmate as a golang library in my project and, in order to test all migrations, I want to apply all rollback migrations until the original state.
To achieve this, I'm using the following function:
```go
func rollbackAll(mate *dbmate.DB) error {
	for {
		err := mate.Rollback()
		if err == nil {
			continue
		}
		// no more migration to rollback
		if err.Error() == "can't rollback: no migrations have been applied" {
			return nil
		}
		if err != nil {
			return err
		}
	}
}
```

In order to avoid issues with future upgrades, would it be possible to export the following error as a global ?

https://github.com/amacneil/dbmate/blob/1613f72c4f9456ebe1b72eb1fa8a55c648a7587a/pkg/dbmate/db.go#L483

like:

```go
const ErrNoRollback = errors.New("can't rollback: no migrations have been applied")
```

I can make a PR if that sounds good to you.

This issue is linked to https://github.com/amacneil/dbmate/issues/221 and https://github.com/amacneil/dbmate/pull/256 but it allows more flexibility when using it as a library.

Thanks!

amacneil:
Hi! Yes that seems like a good idea - PR welcome

gnuletik:
Great ! PR done.

I can also implement the `RollbackAll()` func in a separate PR if you want.

amacneil:
Yeah you could make a PR for rollbackAll too. Curious what you use it for? Why don’t you just truncate or drop all tables?

pawndev:
> Great!
> 
> I'm using it in a go test which:
> 
>     * Creates a temporary database and run all migrations up
> 
>     * Rollback all migrations to the initial state
> 
>     * Re-run all migrations up
> 
> 
> It helps catching missing statements in `migrate:down` blocks.
> 
> For example, if there is a missing `DROP TABLE`, the last up function will fails with an error like `TABLE "something" already exists`.

Not a bad idea. I'll do that in my project too.
Thank you !

gnuletik:
Happy to help :)

For the reference, I created the PR for the RollbackAll func : https://github.com/amacneil/dbmate/pull/303

PR #303, opened 2022-05-31
feat: add RollbackAll to roll back all migrations

gnuletik:


gnuletik:
@amacneil is there anything blocking this PR that I can do? Thanks!

amacneil:
I think `dbmate rollback --all` is a good suggestion.

More thoughts in [this comment](https://github.com/amacneil/dbmate/pull/438#issuecomment-1545258004).

dossy:
Just wondering, @gnuletik, what is the semantic difference between `rollback --all` and `drop` followed by `create`?  And, if they are, in theory, both meant to leave the system in an identical state, wouldn't `drop` followed by `create` be preferred as it's certainly faster to execute than a `rollback --all` would be?

Is there an expectation that `rollback --all` should leave the system in a different state than a `drop` followed by a `create`?  I suppose this is the case if some migrations lack a "migrate down".  But, then the system in that state may not be able to migrate up without error, which is a problem.

gnuletik:
@dossy I totally agree with you.

Rolling back all migrations should leave the database in the same state as a new database if all up/down migrations file are clean.

If you missed something in your down migration, then applying all up migrations will fail. This is the use case I'm trying to cover with this command.

Having a rollback --all command would allow easy testing (e.g. in CI) of the down and up migrations.

dossy:
> Having a rollback --all command would allow easy testing (e.g. in CI) of the down and up migrations.

😲 🤯

Okay, that's a super clever use case for `rollback --all`!

You do a `migrate`, then a `rollback --all`, followed by another `migrate` and if that throws an error the second time around, you know there's an issue with your migration scripts, somewhere.

Well, I'm sold :) on the idea!  I'm sorry I never saw #292 and [your previous explanation in this comment](https://github.com/amacneil/dbmate/issues/292#issuecomment-1142252718) -- thanks for repeating it here.

We'll need to add the `--all` subcommand option to the `rollback` command, update the README, and add some tests, but otherwise this looks good so far.

wryanzimmerman:
Testing all down-migrations in CI is exactly the use-case which lead me to this thread, I'd love to see this added!

wryanzimmerman:
In case it helps anyone here's my intermediate solution:

```
for i in db/migrations/*
do
  if test -f "$i"
  then
    dbmate down
  fi
done;
```

PR #304, opened 2022-06-02
Clickhouse ssl support

shabashev-ivan:


amacneil:
Will need some tests before merging this

pkg/driver/clickhouse/clickhouse.go
```diff
@@ -47,6 +49,33 @@ func connectionString(initialURL *url.URL) string {
 	u.Host = host
 
 	query := u.Query()
+
+	certPath := query.Get("cert_path")
```
amacneil, line 53: Is this a standard parameter used by other ClickHouse clients? I want to avoid dbmate-specific configuration where possible.

amacneil:
I'm not convinced about adding a new url parameter unless this is accepted elsewhere. Please reopen if that can be addressed and test cases provided. Thanks.

PR #306, opened 2022-06-15
Adds flyway to the comparison table

mwhitm:
Resolves https://github.com/amacneil/dbmate/issues/249 for flyway

Some notes which could change the ticks:
- `Timestamp-versioned migration files` I checked this one because a developer can put the timestamp in their migration file name (this was the SOP for the team when I used flyway).  Timestamp versions is not done as default/automatically.  Additionally, flyway also has an `installed_on` column in their migration table.  So dates can be tracked multiple ways.
- `Automatically load .env file` Flyway automatically loads its config file, not sure if this is the same thing.
- `No separate configuration file` there is a config file, however a developer could choose to write a script using only environmental variables
- `Language/framework independent` can be Java dependent, but has a command-line option making it language and platform agnostic.


amacneil:
thanks!

Issue #307, opened 2022-06-21
Add support for new default mysql authentication: caching_sha2_password


shames0:
When trying to use the dbmate container against a mysql > 8.0 compliant server this error is given whenever a schema dump is requested:

```
docker run --env-file=dev.env --rm -it --network=my_network -v $PWD/db:/db ghcr.io/amacneil/dbmate dump
Error: mysqldump: Got error: 1045: "Plugin caching_sha2_password could not be loaded: Error loading shared library /usr/lib/mariadb/plugin/caching_sha2_password.so: No such file or directory" when trying to connect
```

In the case of `up` requests, the `schema.sql` file is just never written to (fails silently) -- this is perhaps a separate issue to be addressed...

It appears the the missing `caching_sha2_password` plugin can be resolved with a simple addition to the `Dockerfile`

```
RUN apk add mariadb-connector-c
```

PR #308, opened 2022-06-21
Include `mariadb-connector-c` (#307)

shames0:
This makes it so that the `caching_sha2_password `  plugin is included so connections to Mysql > 8.0 can succeed.

fixes: #307

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN make build
 FROM alpine as release
 RUN apk add --no-cache \
 	mariadb-client \
-    mariadb-connector-c \
+	mariadb-connector-c \
 	postgresql-client \
 	sqlite \
 	tzdata
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN make build
 FROM alpine as release
 RUN apk add --no-cache \
 	mariadb-client \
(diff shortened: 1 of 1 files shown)


shames0:
This intends to address: #307

amacneil:
Can you explain what this is doing? I'm not familiar with `mariadb-connector-c`.

shames0:
Yes, happy to hash out any questions or concerns here to make sure this is the right path forward.

But the intention with this change is to allow the `docker run ...` usage of `dbmate` (as documented in [the README](https://github.com/amacneil/dbmate/blob/main/README.md?plain=1#L75-L89)) compatible with [MySQL 8 servers using the default/preferred Authentication plugin](https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password). 

As outlined in #307 the `dump` action fails when run against a MySQL 8 server due to the missing `caching_sha2_password` plugin. I was able to find on the alpine package repository [`mariadb-connector-c`](https://pkgs.alpinelinux.org/contents?branch=edge&name=mariadb-connector-c&arch=x86&repo=main) which includes the needed `caching_sha2_password.so` library.

amacneil:
Thanks for the extra info!

PR #316, opened 2022-09-05
Add --number flag on dbmate down

mg98:
It bugged me that I had to rollback migrations one by one if I needed to rollback multiple.
Also, the option to rollback all migrations at once (instead of `drop`) could be useful for CI testing.


gregwebs:
This would interact poorly with #434 (probably throw an error if both are specified).  It seems like this could be accomplished externally with a loop (not sure how annoying that is in your case).

gregwebs:
#256 also tries to tackle this by specifying a target to migrate down to. I think this approach could be safer and avoid accidental off by 1 errors, etc?

mg98:
I'm currently not using this lib in any project so I'm not opinionated. I feel like such an option would be useful and more convenient than typing the version, especially when working collaboratively or on different branches. I don't see this being at particularly high risk for off-by-1 errors.

With regards to #434, you mean when both arguments were to be specified at once? Not sure, maybe.
Might take a look on it again depending how this discussion concludes ;)

gregwebs:
I opened https://github.com/amacneil/dbmate/pull/438 which allows a syntax for specifying a range of migration. If you wanted to specify migrating down the last 2 versions, you could use `-m -2...`. But that needs to be combined with a flag or command that would allow for rolling back multiple migrations at once.

Issue #273, opened 2022-02-16
Feature: Nested directories in migration directory?


JanmanX:
In a project with a lot of migration files, it would be great to structure those in separate sub-directories, such as:
```
├── 0001.sql
├── 0005.sql
├── week1
│   └── 0002.sql
└── week2
    ├── 0003.sql
    └── 0004.sql
```

I know that it is possible to specify the migration directory, but this makes the other folders "invisible" to DBMate. While this still works and won't (seem to) break DBMate, it would be great to have this work natively.

I would like to volunteer to implement this feature, if needed.

rohitpaulk:
I think this is a good idea, don't see a downside to it as long as we're clear that directory naming doesn't affect the order (i.e. files are still ordered by filenames irrespective of what directory they're in). 

@amacneil thoughts?

amacneil:
Only potential source of confusion I can see:

Would you expect the migrations to be applied in alphabetical order including directory name, or ignore the directory and only look at the version number?

JanmanX:
Personally I would prefer applying migrations using the version number, regardless which directory they come from.
The advantages I can think of:
- More freedom how to structure the sub-directories. You could structure the migrations in directories: `functions`, `tables`, `roles`, etc., and not worry which order they are applied in.
- The migration versions stay the same and can further moved around without worrying of breaking the ordering.
- It might be more intuitive to figure out which order to apply migrations in (less complexity)

amacneil:
My fear is that another group of people will expect the opposite behavior - if you have folders named "week1", "week2", etc - why wouldn't they be applied in order?

Likewise if people organized their migrations into version number folders - "v1.0", "v1.1", etc. Or even worse, they might expect us to respect semver ordering ("v1.23" folder being applied after "v1.5").

On a tangent: if you have a lot of migrations (e.g. project that has been under development for years), you can always consolidate your old migrations. Just take the current state of the database and put it in a new migration file (making sure all the DDL statements use "if exists" where appropriate so that they won't make any actual changes to existing databases), then delete all the old migration files.

CasperLindinger:
What if you added a command like `dbmate new-dir week1`, which adds a version number to a directory like the migrations?

```
├── 0001.sql    
├── 0005.sql    
├── 0001_week1  
│   └── 0002.sql
└── 0002_week2  
    ├── 0003.sql
    └── 0004.sql
```

Then perhaps it's matter of starting with the root directory and then go through the version numbers on the directories

[...]

PR #336, opened 2022-10-31
Enable nested directories in migration directory

iepsenn:
Uses the idea discussed in pr https://github.com/amacneil/dbmate/issues/273, allowing separate migrations into subdirectories.


amacneil:
Thank you for the PR, but as discussed in https://github.com/amacneil/dbmate/issues/273 I don't see a need for this feature and I feel it introduces unnecessary sources of confusion (which order should migrations be applied in, what happens if there are conflicting version numbers in different folders, etc).

Issue #379, opened 2023-01-12
ClickHouse CREATE DATABASE IF NOT EXISTS syntax is incorrect


farbodsalimi:
**Description**

I wasn't sure if this was a known issue or not so that's why I didn't create a PR. The `CREATE DATABASE` statement in https://github.com/amacneil/dbmate/blob/cd19cb4f672e0804c6ab6204287f170aea5d9b44/pkg/driver/clickhouse/clickhouse.go#L146 looks incorrect. According to https://clickhouse.com/docs/en/sql-reference/statements/create/database/ I think we should replace that line with:

```go
 buf.WriteString("CREATE DATABASE IF NOT EXISTS" + drv.quoteIdentifier(databaseName) + ";\n\n") 
```

- Version: 1.16.0
- Database: ClickHouse
- Operating System: macOS Ventura

**Steps To Reproduce**

N/A

**Expected Behavior**

N/A

amacneil:
Looks like you are right. ClickHouse is community maintained, happy to accept a PR fixing it.

amacneil:
Fixed in #380

PR #380, opened 2023-01-29
Fix create db and upgrade clickhouse-server to v22.8

farbodsalimi:
- Upgraded `clickhouse-server` image to v22.8 the latest LTS version
  - Updated the tests since the error messages are different in v22
- Fixed `CREATE DATABASE IF NOT EXISTS` based on the official document https://github.com/amacneil/dbmate/issues/379


amacneil:
Thanks!

PR #383, opened 2023-02-17
SQLite: Do not dump system tables

### LINKED ISSUE #315: Invalid Sqlite schema.sql when INTEGER PRIMARY KEY AUTOINCREMENT is used
## TL;DR 
 
Dbmate always outputs the full schema, including `sqlite_sequence` which is auto-generated when autoincrement appears in at least one table definition. This table is reserved for internal use in sqlite and cannot be created by the user (see https://www.sqlite.org/autoinc.html section 3).  

This schema.sql file fails to load on an empty database. 

## Setup: 
dbmate 1.15.0, sqlite 3.39 

## To reproduce:

Run `dbmate new test`, let's call it `db/migrations/20220818223405_test.sql`.

The file should contain:

```
-- migrate:up

create table t (
  id INTEGER PRIMARY KEY AUTOINCREMENT
);

-- migrate:down

drop table t
```

Run `dbmate -u sqlite3:test.db up`

The file `db/schema.sql` should contain: 

```
CREATE TABLE IF NOT EXISTS "schema_migrations" (version varchar(255) primary key);
CREATE TABLE t (
  id INTEGER PRIMARY KEY AUTOINCREMENT
);
CREATE TABLE sqlite_sequence(name,seq);
-- Dbmate schema migrations
INSERT INTO "schema_migrations" (version) VALUES
  ('20220818223405');
```

Now test the schema file, it should be possible to use on an empty db to initialise it.

Run `sqlite3 test2.db`

```
SQLite version 3.39.....
Enter ".help" for usage hints.
sqlite> .read db/schema.sql
Parse error near line 5: object name reserved for internal use: sqlite_sequence
```

## Solution 

The table `sqlite_sequence` should not be extracted in the schema, sqlite will generate it automatically.

docapotamus:
Pass the `--nosys` flag through to the SQLite `.schema` command. This will stop the `sqlite` command from writing out system tables `sqlite_*`.

This flag is documented:

```
sqlite> .help .schema
.schema ?PATTERN?        Show the CREATE statements matching PATTERN
   Options:
      --indent             Try to pretty-print the schema
      --nosys              Omit objects whose names start with "sqlite_"
```

I wasn't sure if this would constitute a breaking change so I didn't hide it behind a flag as the issue was marked as `Bug`, but I'm happy to add this if it is correct to.

This was also quite hard to test as the test suite passed with or without the `--nosys` flag. I'm presuming this is a change in SQLite but I couldn't find a reference to it. I added a test any way just to make sure it doesn't creap back in.

I also confirmed this behaviour on the command line.

Fixes #315

### DIFF
--- a/pkg/driver/sqlite/sqlite.go
+++ b/pkg/driver/sqlite/sqlite.go
@@ -131,7 +131,7 @@ func (drv *Driver) schemaMigrationsDump(db *sql.DB) ([]byte, error) {
 // DumpSchema returns the current database schema
 func (drv *Driver) DumpSchema(db *sql.DB) ([]byte, error) {
 	path := ConnectionString(drv.databaseURL)
-	schema, err := dbutil.RunCommand("sqlite3", path, ".schema")
+	schema, err := dbutil.RunCommand("sqlite3", path, ".schema --nosys")
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/driver/sqlite/sqlite_test.go
+++ b/pkg/driver/sqlite/sqlite_test.go
@@ -180,9 +180,14 @@ func TestSQLiteDumpSchema(t *testing.T) {
 	err = drv.InsertMigration(db, "abc2")
 	require.NoError(t, err)
 
+	// create a table that will trigger `sqlite_sequence` system table
+	_, err = db.Exec("CREATE TABLE t (id INTEGER PRIMARY KEY AUTOINCREMENT)")
+	require.NoError(t, err)
+
 	// DumpSchema should return schema
 	schema, err := drv.DumpSchema(db)
 	require.NoError(t, err)
+	require.Contains(t, string(schema), "CREATE TABLE t (id INTEGER PRIMARY KEY AUTOINCREMENT)")
 	require.Contains(t, string(schema), "CREATE TABLE IF NOT EXISTS \"test_migrations\"")
 	require.Contains(t, string(schema), ");\n-- Dbmate schema migrations\n"+
 		"INSERT INTO \"test_migrations\" (version) VALUES\n"+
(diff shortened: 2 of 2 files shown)


amacneil:
Can you tell why the CI test is failing? It's not obvious to me.

docapotamus:
I'm not to sure. Looks to be the same issue in the previous PR. I'll try take a look later today

amacneil:
Rebase should fix it

pkg/driver/sqlite/sqlite.go
```diff
@@ -131,7 +131,7 @@ func (drv *Driver) schemaMigrationsDump(db *sql.DB) ([]byte, error) {
 // DumpSchema returns the current database schema
 func (drv *Driver) DumpSchema(db *sql.DB) ([]byte, error) {
 	path := ConnectionString(drv.databaseURL)
-	schema, err := dbutil.RunCommand("sqlite3", path, ".schema")
+	schema, err := dbutil.RunCommand("sqlite3", path, ".schema --nosys")
```
amacneil, line 134: Pretty sure these should be separate arguments right?
```suggestion
	schema, err := dbutil.RunCommand("sqlite3", path, ".schema", "--nosys")
```

Are you sure this change did anything for you? Since you mentioned the test never failed, in which situations did you see those tables get created?

docapotamus: I can recreate this using the steps in the issue just not in the unit test.

If I add the `.schema --nosys` as one string the `sqlite_sequence` does not get dumped to the schema. If I add it as a sepeate string I receive an error.

#### No change applied

```sh
j.doherty@test my_dbmate % DATABASE_URL=sqlite:db/db.sqlite ./dist/dbmate dump
Writing: ./db/schema.sql
j.doherty@test my_dbmate % cat ./db/schema.sql 
CREATE TABLE IF NOT EXISTS "schema_migrations" (version varchar(255) primary key);
CREATE TABLE t (
      id INTEGER PRIMARY KEY AUTOINCREMENT
);
CREATE TABLE sqlite_sequence(name,seq);
-- Dbmate schema migrations
INSERT INTO "schema_migrations" (version) VALUES
  ('20230220094634');
```

#### With my patch applied

```sh
j.doherty@test my_dbmate % DATABASE_URL=sqlite:db/db.sqlite ./dist/dbmate dump
Writing: ./db/schema.sql
j.doherty@test my_dbmate % cat ./db/schema.sql                                
CREATE TABLE IF NOT EXISTS "schema_migrations" (version varchar(255) primary key);
CREATE TABLE t (
      id INTEGER PRIMARY KEY AUTOINCREMENT
);
-- Dbmate schema migrations
INSERT INTO "schema_migrations" (version) VALUES
  ('20230220094634');
```

#### With the flag in a seperate arg

```sh
j.doherty@test my_dbmate % DATABASE_URL=sqlite:db/db.sqlite ./dist/dbmate dump
Error: sqlite3: Error: unknown option: -nosys
Use -help for a list of options.
```

amacneil: I see. That table only shows up in the dump if you have an `AUTOINCREMENT` field. I added one to the test and it correctly fails now.

docapotamus: Good spot. Don't know why I didn't think to check the test migration.

PR #384, opened 2023-02-18
Fix: Docker Test CI errors

docapotamus:
I am trying to resolve the VCS tagging issue in the CI pipline.

The change in Go is documented here: https://go.dev/doc/go1.18. I am unsure why this is failing in the pipeline when it works locally running the same commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN curl -fsSL https://raw.githubusercontent.com/golangci/golangci-lint/master/i
 COPY go.* /src/
 RUN go mod download
 COPY . /src/
+ENV GO_ADDITIONAL_FLAGS="-buildvcs=false"
 RUN make build
 
 # release stage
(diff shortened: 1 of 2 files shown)


docapotamus:
Passing `-buildvcs=false` is incorrect in the pipeline as that uses Go 1.17 where this isn't available.

The Dockerfile uses Go 1.19 where this is available and causes an error. I presume because there is some issue with it inside a Docker build

docapotamus:
I presume the actual issue has something to do with permissions to the .git folder but I can't be sure. All steps recreated on my Mac where correct.

Without changing to much I allowed adding additional build FLAGS and in the case of the Docker test disabled `buildvcs`

Not sure if this is the best solution but it works. Ideally the Go compiler would do the right thing but with the test being via Docker compose I wasn't sure what was best.

@amacneil any thoughts?

amacneil:
Thanks for the links. Unclear why this started failing now - I guess due to an updated git binary somehow.

I figured out a solution that maintains the vcs information (https://github.com/amacneil/dbmate/pull/385).

PR #387, opened 2023-02-20
Clickhouse: Fix CheckMigrationTableExists

### LINKED ISSUE #301: `dbmate status` fails for ClickHouse
greetings!

context:
- `ClickHouse server version 22.3.3 revision 54455.`
- `dbmate status` fails at `1.15.0` but works at `1.13.0`

```sh
$ dbmate --url <clickhouse_db_url> status
Error: sql: expected 0 arguments, got 1
```

based on my quick local debug, I think [this function](https://github.com/amacneil/dbmate/blob/d671075f1fbce82dffa3541abc2aacaded4d200d/pkg/driver/clickhouse/clickhouse.go#L233-L243) is the culprit, which was introduced in the [`1.14.0`](https://github.com/amacneil/dbmate/releases/tag/v1.14.0)

thanks

docapotamus:
Checking for the migration tables existence in Clickhouse shouldn't pass a parameter.

This could be acheived one of two ways:

- Use Sprintf to add the table name to the query (this is what is used elsewhere within the Clickhouse driver)
- Check the `information_schema` using `currentDatabase()`.

Opted for the first as that is what is used elsewhere.

Fixes #301

### DIFF
--- a/pkg/driver/clickhouse/clickhouse.go
+++ b/pkg/driver/clickhouse/clickhouse.go
@@ -233,7 +233,7 @@ func (drv *Driver) DatabaseExists() (bool, error) {
 // MigrationsTableExists checks if the schema_migrations table exists
 func (drv *Driver) MigrationsTableExists(db *sql.DB) (bool, error) {
 	exists := false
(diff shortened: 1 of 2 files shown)


amacneil:
thanks!

Issue #403, opened 2023-02-27
Absolute path for migration directory is no longer working


blinktag:
v2.0.0 broke the ability to specify an absolute path for the migrations directory.

The culprit is likely https://github.com/amacneil/dbmate/pull/388/files#diff-85e7f15f76b0a01f7ef13c1a5e83e93ba61401694dc5118dba4180cc8f4d0a59R73 which will cause any path to be prefixed with the argument passed to `DirFS()`

- Version: v2.0.0
- Database: postgres
- Operating System: alpine

**Steps To Reproduce**

The following command works correctly with v1.16.2:

```
$ dbmate -d /migrations migrate
```

Upgrading to v2.0.0 causes the following error:

```
Error: could not find migrations directory `/migrations`
```

**Expected Behavior**

Absolute paths should work as with previous versions

amacneil:
Derp

PR #406, opened 2023-02-28
Fix absolute migration directory support

amacneil:
Closes https://github.com/amacneil/dbmate/issues/403

### DIFF
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -39,7 +39,7 @@ type DB struct {
 	AutoDumpSchema bool
 	// DatabaseURL is the database connection string
 	DatabaseURL *url.URL
-	// FS allows overriding the filesystem
+	// FS specifies the filesystem, or nil for OS filesystem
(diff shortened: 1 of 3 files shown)


nielslanting:
I think this commit broke Windows paths when working with the Go embed package. I'm trying to use dbmate as library as specified in https://github.com/amacneil/dbmate#use-dbmate-as-a-library.

`readMigrationsDir` is called with parameter dir `./db/migrations`.
`readMigrationsDir` uses `filepath.Clean`.
I assume on Linux machines this will turn it into: `db/migrations` while on Windows this will get turned into `db\\migrations`.

https://pkg.go.dev/path/filepath#Clean has different behaviour on Windows.
> On Windows, Clean does not modify the volume name other than to replace occurrences of "/" with `\`. For example, Clean("//host/share/../x") returns `\\host\share\x`.

 `db\\migrations` is not a valid path for the embedded fs. So it's unable to read the files/dir from the embedded fs.

PR #412, opened 2023-03-03
Port JS library from defunctzombie/node-dbmate

defunctzombie:
Add a library interface to the dbmate npm package. This interface allows developers to invoke `dbmate` from their node programs. A common usecase is to invoke dbmate commands during unit testing.

I tested this change by using `npm pack ./packages/dbmate` after running generate. This creates a `.tar.gz` file I can install using `yarn` or `npm`.


typescript/generate.ts
```diff
@@ -113,6 +114,10 @@ async function main() {
   // copy readme and license
   await copyFile("../LICENSE", "packages/dbmate/LICENSE");
   await copyFile("../README.md", "packages/dbmate/README.md");
+
+  await exec("npm", ["run", "build"], {
+    cwd: "packages/dbmate",
+  });
```
defunctzombie, line 120: Another location for this could be the CI steps rather than here in `generate.ts`

defunctzombie:
@amacneil PTAL

amacneil:
Partially landed in https://github.com/amacneil/dbmate/pull/418

defunctzombie:
Closing this out in favor of using `resolveBinary` from the published package and using our own wrapper for now.

PR #414, opened 2023-03-11
Add Sqitch to comparison chart

ttfkam:
References issue https://github.com/amacneil/dbmate/issues/249

Addresses issues in PR https://github.com/amacneil/dbmate/pull/268


amacneil:
Thanks!

FYI I removed "version control aware" since I don't know what that means, and fixed the formatting.

ttfkam:
Version control aware means that the database migration tool is aware that its migration files are being checked into version control and allows for fewer conflicts among team members when more than one is making database migration files at a time. It is a situation that is quite common and one where timestamps are not adequate to handle safely. VCS awareness also allows for bundling and deployment by tag rather than just the most recent commit to the repository.

https://speakerdeck.com/theory/sane-database-change-management-with-sqitch?slide=227

Also allows for editing functions with a real diff, not two unrelated files that happen to hold the same function name. The "where is the latest version of that function/view?" problem. Code reviews become much easier with this.

https://speakerdeck.com/theory/sane-database-change-management-with-sqitch?slide=280

This is why Sqitch doesn't support versions by timestamps. It was not an oversight or missing feature. Just the opposite in fact. It is why I added the row. As I look at it, I should have added a testing/verification row as well, so that migrations can be checked in an automated fashion to be sure they worked as intended rather than just being syntactically correct, which triggers no migration errors in many tools.

Note: none of this is meant as a slight against dbmate. What you've made here is very impressive. I especially like schema export functionality for easier diffing. It mitigates some of the problems with finding the most recent version of an object for update.

amacneil:
Thanks for the extra info. I tried to look through that deck but still can't understand how it handles out of order migrations differently. Anyway, it sounds like too much to convey in the comparison table - it's not meant to be comprehensive, just an answer to "why create dbmate".

The function editing/diff feature in squitch sounds cool.

PR #418, opened 2023-03-13
Convert dbmate npm package to esm/typescript

amacneil:
Convert `dbmate` npm package to typescript and esm, in preparation for adding typescript library.

Extracted from https://github.com/amacneil/dbmate/pull/412


typescript/packages/dbmate/src/resolveBinary.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Resolve path to dbmate for the current platform
+ * */
+export function resolveBinary(): string {
+  const ext = platform === "win32" ? ".exe" : "";
+  const path = `@dbmate/${platform}-${arch}/bin/dbmate${ext}`;
+
+  try {
+    return createRequire(import.meta.url).resolve(path);
```
amacneil, line 12: @defunctzombie for future reference, `createRequire()` is the magic needed to access require from esm.

defunctzombie: 😮

PR #432, opened 2023-04-11
ClickHouse: Support migrations on cluster

FatherCandle:
Adding functionally to support usage of Dbmate with ClickHouse cluster.

Easy default usage with just supplying "on_cluster" query parameter.

Took the approach of trying to keep the changes simple without breaking any existing stuff.
95% of the changes in the PR are for the tests and setups of the ClickHouse cluster dockers.

Will happily answer any questions, and would really appreciate reviews.

Solves existing requests in discussions:
https://github.com/amacneil/dbmate/discussions/356
https://github.com/amacneil/dbmate/discussions/244


pkg/driver/clickhouse/clickhouse_cluster_test.go
```diff
@@ -0,0 +1,351 @@
...
+	err = db.Ping()
+	if shouldExist {
+		require.NoError(t, err)
+	} else {
+		require.EqualError(t, err, "code: 81, message: Database dbmate_test doesn't exist")
+	}
+}
+
+// Makes sure driver creatinon is atomic
+func TestDriverCreationSanity(t *testing.T) {
```
Mizaro, line 39: Nice additional test. Maybe we should suggest enforcing it for all drivers.

Mizaro (approved):
Nice implementation! LGTM :)

amacneil:
@FatherCandle I gave you commit access, so feel free to merge if there is sufficient consensus on code review + test cases.

@dossy @gregwebs

FatherCandle:
> @FatherCandle I gave you commit access, so feel free to merge if there is sufficient consensus on code review + test cases.
> 
> @dossy @gregwebs

That's awesome :) 
I would appreciate the CR, it seems scary because I had to add manually Clickhouse config files for the tests.
But the actual changes aren't that big.

pkg/driver/clickhouse/clickhouse.go
```diff
@@ -250,15 +266,20 @@ func (drv *Driver) MigrationsTableExists(db *sql.DB) (bool, error) {
 
 // CreateMigrationsTable creates the schema migrations table
 func (drv *Driver) CreateMigrationsTable(db *sql.DB) error {
+	engineClause := "ReplacingMergeTree(ts)"
+	if drv.clusterParameters.OnCluster {
+		engineClause = fmt.Sprintf("ReplicatedReplacingMergeTree('%s', '%s', ts)", drv.clusterParameters.ZooPath, drv.clusterParameters.ReplicaMacro)
```
dossy, line 271: Here's another place where a `drv.escapeString()` could be used.

pkg/driver/clickhouse/clickhouse.go
```diff
@@ -95,6 +99,14 @@ func (drv *Driver) openClickHouseDB() (*sql.DB, error) {
 	return sql.Open("clickhouse", clickhouseURL.String())
 }
 
+func (drv *Driver) onClusterClause() string {
+	clusterClause := ""
+	if drv.clusterParameters.OnCluster {
+		clusterClause = fmt.Sprintf(" ON CLUSTER '%s'", drv.clusterParameters.ClusterMacro)
```
dossy, line 105: It's unfortunate that we can't use `drv.quoteIdentifier()` here, because these are strings and not identifiers, which are quoted/escaped differently.

It is probably worth introducing a new `drv.escapeString()` function that escapes `'` as `\'` and `\` as `\\`, per [the ClickHouse documentation on strings](https://clickhouse.com/docs/en/sql-reference/syntax#string), and use that here, and anywhere else we have string values that need escaping.

dossy (changes requested):
This all looks great!

Would you mind also updating [the ClickHouse section of the README](README.md#clickhouse) with documentation listing the 4 new connection query params this PR introduces (`on_cluster`, `zoo_path`, `cluster_macro`, and `replica_macro`), too, please?

pkg/driver/clickhouse/clickhouse.go
FatherCandle: I have added drv.escapeString method + a test for it.

I liked it more as escaping method then a quoting method, so I made the return value be the escaped string, while keeping the '' around the %s for it to be explicit.

pkg/driver/clickhouse/clickhouse.go
FatherCandle: Used the newly created drv.escapeString method.

FatherCandle:
> This all looks great!
> 
> Would you mind also updating [the ClickHouse section of the README](https://github.com/FatherCandle/dbmate/blob/chore/support_clickhouse_cluster/README.md#clickhouse) with documentation listing the 4 new connection query params this PR introduces (`on_cluster`, `zoo_path`, `cluster_macro`, and `replica_macro`), too, please?

@dossy 
Added the docs in the README section of ClickHouse + the escape string suggetsion.
You are welcome to review it again :)

dossy (approved):
The changes look great, thanks!  I think the documentation will be particularly useful for anyone who wants to take advantage of this new functionality.

PR #434, opened 2023-04-20
specify just a single migration to use

gregwebs:
It's not unusual for there to be multiple pending migrations but to want to carefully apply just one.
I have been forced to engage in workarounds like deleting the migration files I don't want to apply (probably most people are putting the migration file they want into a separate directory).

This was tested manually with `dbmate status`  by using the version and the filename and seeing that only one migration showed up. Also tested by using a non-existent version and seeing the error show up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ The following options are available with all commands. You must use command line
 
 - `--url, -u "protocol://host:port/dbname"` - specify the database url directly. _(env: `DATABASE_URL`)_
 - `--env, -e "DATABASE_URL"` - specify an environment variable to read the database connection URL from.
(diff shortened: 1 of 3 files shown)


gregwebs:
This PR is updated to allow for multiple migrations to be specified rather than just one.

gregwebs:
#256 for automatically applying multiple migrations at once by specifying one does not satisfy my workflows: I would still have to engage in workarounds. However, someone that wants that functionality can use this PR to apply multiple migrations in one go.

It would be possible to add a special syntax to this MR to directly support what is desired in #256. Something like adding a star in front: `*migration` would signify that all the migrations up to that one should be applied as well. The nice think about this PR is that it works with `dbmate status`. So you could use `*migration` with status to see ahead of time what would get applied. I love tools that can tell me exactly what they will do ahead of time.

PR #435, opened 2023-04-21
Postgres: support a connection string DSN

gregwebs:
use DATABASE_DSN and DBMATE_DRIVER
The format of DATABASE_DSN is defined under connection string here: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

The DATABASE_URL is problematic because it is parsed to a URL in Go. This requires encoding the password for the URL which does not necessarily get decoded properly.

It is simpler to not alter the parameters to fit them into a url: that is what DATABASE_DSN does. Currently it will fail if a configuration parameter has a space. However, spaces are normally not used for any of the parameters. And that limitation can be fixed by taking on a DSN parser dependency.

## Testing

Unit tests for DSN manipulation.
Manual tests were done to see that DATABASE_DSN works with `dbmate status`. I am using it for other commands now as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,8 @@ $ dbmate -u "postgres://postgres@127.0.0.1:5432/myapp_test?sslmode=disable" up
 
 The only advantage of using `dbmate -e TEST_DATABASE_URL` over `dbmate -u $TEST_DATABASE_URL` is that the former takes advantage of dbmate's automatic `.env` file loading.
 
+Some drivers (currently just Postgres) support specifying a DSN string instead of a url by using `DATABASE_DSN` instead of `DATABBASE_URL`. When using `DATABASE_DSN` you can specify the driver with `DBMATE_DRIVER`.
+
 #### PostgreSQL
 
 When connecting to Postgres, you may need to add the `sslmode=disable` option to your connection string, as dbmate by default requires a TLS connection (some other frameworks/languages allow unencrypted connections by default).
--- a/main.go
+++ b/main.go
@@ -37,7 +37,7 @@ func NewApp() *cli.App {
 	app.Usage = "A lightweight, framework-independent database migration tool."
 	app.Version = dbmate.Version
 
-	defaultDB := dbmate.New(nil, nil)
+	defaultDB := dbmate.New(nil)
 	app.Flags = []cli.Flag{
 		&cli.StringFlag{
 			Name:    "url",
@@ -230,7 +230,7 @@ func action(f func(*dbmate.DB, *cli.Context) error) cli.ActionFunc {
 		if err != nil {
 			return err
 		}
-		db := dbmate.New(u, dsn)
+		db := dbmate.NewWithDSN(u, dsn)
 		db.AutoDumpSchema = !c.Bool("no-dump-schema")
 		db.MigrationsDir = c.String("migrations-dir")
 		db.MigrationsTableName = c.String("migrations-table")
--- a/main.go
+++ b/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/urfave/cli/v2"
 
 	"github.com/amacneil/dbmate/v2/pkg/dbmate"
+	"github.com/amacneil/dbmate/v2/pkg/dbutil"
 	_ "github.com/amacneil/dbmate/v2/pkg/driver/clickhouse"
 	_ "github.com/amacneil/dbmate/v2/pkg/driver/mysql"
 	_ "github.com/amacneil/dbmate/v2/pkg/driver/postgres"
@@ -36,7 +37,7 @@ func NewApp() *cli.App {
 	app.Usage = "A lightweight, framework-independent database migration tool."
 	app.Version = dbmate.Version
 
-	defaultDB := dbmate.New(nil)
+	defaultDB := dbmate.New(nil, nil)
 	app.Flags = []cli.Flag{
 		&cli.StringFlag{
 			Name:    "url",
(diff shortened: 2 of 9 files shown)


gregwebs:
@dossy thanks for attending to pull requests. I am still working on this pull request.

At a high level it seems that the point of this pull request is still not understood.
I don't see the point in doing a code review if the purpose is not understood.
As per the description:

> The DATABASE_URL is problematic because it is parsed to a URL in Go. This requires encoding the password for the URL which does not necessarily get decoded properly.

I had an auto-generated (strong) db password that did not work due to the URL encoding issues (yes I did try encoding it). Whereas the key=value connection string just works. So parsing it into a URL would defeat the purpose. Another approach could be to not parse the url but just leave it as a string (Postgres doesn't actually want it to be url encoded), but this would be a big change that would require changing every driver. And it is still worse than key=value because '@' still needs to get escaped in a password. Whereas I am quite happy to not put any spaces in my configuration.

amacneil:
I don't think I understand the problem this PR is solving. You need to encode special characters in the password using URL encoding, that should be very simple in any programming language. Or you could instead generate a random strong password using URL-safe characters. Either option is fine.

Any connection string format will need to escape special characters (e.g. if your password contains a space such as `3jrk foo=bar` will it break your parser implementation?).

If you are having problems with URL encoding please post details and we can help you.

dossy:
> [...] I am still working on this pull request.

If this pull request isn't ready for review, I suggest you mark it as such by [converting it to a draft pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests).

> I had an auto-generated (strong) db password that did not work due to the URL encoding issues (yes I did try encoding it). Whereas the key=value connection string just works. So parsing it into a URL would defeat the purpose. [...]

I would appreciate seeing an example password that cannot be used by dbmate when URL-encoded correctly.

As an experiment, I gave this a try:

```shell
# start a PostgreSQL 10 test server
$ docker run --rm --name postgres-test -d -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres:10
b0ce2a7d9cf4d7b80e5721e1ff5e32a82c9108a1c580d68b5cb5122a5b95ba94

# generate a password of all printable ASCII characters
$ printables=$(i=33; while [ $i -le 126 ]; do printf "\x$(printf %x $i)"; test $i -eq 79 && printf " "; i=$(( i + 1 )); done)

# confirm that our password contains what we think it does
$ echo $printables
!"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNO PQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~

# encrypt the password for PostgreSQL
$ printf "%s%s" "$printables" test | md5sum
16d8f2a726c35fe08c61fe9c780c6562  -

# create our test user with our all-printable-ASCII password
$ PGPASSWORD=postgres psql -h localhost -p 5432 -U postgres -c "CREATE USER test WITH ENCRYPTED PASSWORD 'md516d8f2a726c35fe08c61fe9c780c6562';"
CREATE ROLE

# confirm that we can connect to our test user with psql
$ PGPASSWORD="$printables" psql -h localhost -p 5432 -U test -d postgres -c "SELECT current_user;"
 current_user 
--------------
 test
(1 row)

# let's try connecting using database URL from dbmate with an incorrect password
$ docker run --rm --name dbmate --link postgres-test:db amacneil/dbmate --url "postgres://test:test@db:5432/postgres?sslmode=disable" status
Error: pq: password authentication failed for user "test"

# now, let's encode the password
$ export printables
$ node -e 'process.stdout.write(process.env.printables);'
!"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNO PQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~
$ encoded="$(node -e 'process.stdout.write(encodeURIComponent(process.env.printables));')"
$ echo $encoded
!%22%23%24%25%26'()*%2B%2C-.%2F0123456789%3A%3B%3C%3D%3E%3F%40ABCDEFGHIJKLMNO%20PQRSTUVWXYZ%5B%5C%5D%5E_%60abcdefghijklmnopqrstuvwxyz%7B%7C%7D~

# and try again
$ docker run --rm --name dbmate --link postgres-test:db amacneil/dbmate --url "postgres://test:$encoded@db:5432/postgres?sslmode=disable" status
Error: could not find migrations directory `./db/migrations`

# okay, let's create a migration
$ mkdir -p db/migrations
$ printf -- "-- migrate:up\nselect current_user;\n-- migrate:down\n" > db/migrations/1_test.sql

# and use it
$ docker run --rm --name dbmate --link postgres-test:db --volume $(pwd)/db:/db amacneil/dbmate --url "postgres://test:$encoded@db:5432/postgres?sslmode=disable" status
[ ] 1_test.sql

Applied: 0
Pending: 1

# and run it
$ docker run --rm --name dbmate --link postgres-test:db --volume $(pwd)/db:/db amacneil/dbmate --url "postgres://test:$encoded@db:5432/postgres?sslmode=disable" up --verbose
Applying: 1_test.sql
Rows affected: 1
Writing: ./db/schema.sql
```

dbmate has no problem connecting to PostgreSQL with a password containing any of the printable ASCII characters in it, if the password is properly encoded.

One Go function that parses a ODBC-style connection string and produces a properly encoded database URL is all that's needed.

gregwebs:
I was able to get the password encoding working (in the encoding process I accidentally introduced a newline). It's still a hassle though and key=value would simplify things since it's very rare to want a space, but very common to want a url encoded character in a password.

I will close this PR though since the maintainers don't like it. There are still a lot of other obviously good PRs that need to get merged- better to spend energy there.

dossy:
I do think adding `--connection-string value` and parsing the ODBC-style connection string and returning a well-formed database URL is a good idea, and would provide you the functionality that you were looking for in this PR.

The difficulty here is that, unfortunately, PostgreSQL [connection strings](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) do not follow the [ODBC specification for connection strings](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqldriverconnect-function?view=sql-server-ver16#comments).  😞 

This means having a single connection string parsing implementation won't be possible, and it'll need to be per-driver, and we'll need something like `--driver value` to specify which driver the connection string is intended for.

I captured these details in [a new issue](https://github.com/amacneil/dbmate/discussions/436) so when someone does go to implement connection string support, these details aren't overlooked.

Issue #118, opened 2020-02-03
Environment Variables within SQL Scripts


xemoka:
Hello, 

I would like access to variables within the .env file from each .sql file; i.e.: if I have

```sh
# .env
ROLE="authentication"
ROLE_PASS="hunter2"
```

Then I could use a placeholder of 

```sql
-- 2020020319418_db_role.sql
create role :"ROLE" login password :'ROLE_PASS'
```

Is this of interest to folks or is this already possible and I just can't seem to see how? It doesn't seem like golang's _pq_ support's psql's `\set` type statements in scripts either. Maybe there's a way to do this through _pq_ I'm unaware of?

amacneil:
Interesting idea. I don't think it's possible today.

I'm wondering if there are any clever ways to solve this. Doing basic string interpolation would be possible, but it could lead to some issues with improper escaping depending on whether you are using the variable as a literal string or as an identifier.

xemoka:
In psql `\set variable value` results in `:variable` returning the value, if you need it quoted, you provide `:'variable'` or `:"variable"` and it outputs the quoted variable. 

Does go's _text/template_ provide a way to do this easily based on how .sql files are read into dbmate? I wish I was more versed in golang...

amacneil:
I'm going to close this issue, since it doesn't seem like a critical feature for dbmate, and there doesn't seem to be any reasonable way to implement it.

Problems I can see:
* We don't do any sql parsing, so it's not safe to simply inject variables directly
* I don't think the golang `lib/pq` library we use allows you to set variables using the syntax you mentioned
* Anything we add should probably work cross-platform, and while `psql` appears to allow you to parameterize sql files, I don't think all databases do

Finally, I think you could fairly easily work around the lack of injected variables in dbmate by simply running a post-migration script or one-liner using `psql`.

I'm open to discussing further if someone can suggest an elegant way to do this while addressing the issues above.

amacneil:
After thinking about this some more, I think the easiest way to add it would be to use the Go template library. So your migrations would end up something like this:

```
create role {{ .Role }} login password {{ .RolePass }}
```

and then maybe require that variables are explicitly specified when you invoke dbmate (so that we don't have to inherit/expose the entire environment):


```
$ dbmate up --parameter "Role=$ROLE"
```

We can escape the values as string literals by default, which should be safe for most uses.

[...]

PR #439, opened 2023-05-01
Use system variables in migrations

davidecavestro:
Add support to selectively enable usage of env vars within migrations.
Named env vars are enabled though "env-var:VAR_NAME" entries within the up/down block headers., reusing the existing parsing logic
The name of activated env vars are then kept with migration options and passed to go templating just before executing relevant migrations.

See discussion https://github.com/amacneil/dbmate/discussions/352 and issue https://github.com/amacneil/dbmate/issues/118

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       MYSQL_TEST_URL: mysql://root:root@mysql/dbmate_test
       POSTGRES_TEST_URL: postgres://postgres:postgres@postgres/dbmate_test?sslmode=disable
       SQLITE_TEST_URL: sqlite3:/tmp/dbmate_test.sqlite3
+      YABBA_DABBA_DOO: Yabba dabba doo!
 
   dbmate:
     build:
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -1,6 +1,7 @@
 package dbmate
 
 import (
+	"bytes"
 	"database/sql"
 	"errors"
 	"fmt"
@@ -11,6 +12,8 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	"strings"
+	"text/template"
 	"time"
 
 	"github.com/amacneil/dbmate/v2/pkg/dbutil"
@@ -329,7 +332,12 @@ func (db *DB) Migrate() error {
 
 		execMigration := func(tx dbutil.Transaction) error {
 			// run actual migration
-			result, err := tx.Exec(parsed.Up)
+			upScript, err := db.resolveRefs(parsed.Up, parsed.UpOptions.EnvVars())
+			if err != nil {
+				return err
+			}
+
+			result, err := tx.Exec(upScript)
 			if err != nil {
 				return err
 			} else if db.Verbose {
@@ -359,6 +367,38 @@ func (db *DB) Migrate() error {
 	}
 
 	return nil
+}
+
+func (db *DB) resolveRefs(snippet string, envVars []string) (string, error) {
+	if envVars == nil {
+		return snippet, nil
+	}
+
+	envMap := db.getEnvMap()
+	model := make(map[string]string, len(envVars))
+	for _, envVar := range envVars {
+		model[envVar] = envMap[envVar]
+	}
+
+	template := template.Must(template.New("tmpl").Parse(snippet))
+
+	var buffer bytes.Buffer
+	if err := template.Execute(&buffer, model); err != nil {
+		return "", err
+	}
+
+	return buffer.String(), nil
+}
+
+func (db *DB) getEnvMap() map[string]string {
+	envMap := make(map[string]string)
+
+	for _, envVar := range os.Environ() {
+		entry := strings.SplitN(envVar, "=", 2)
+		envMap[entry[0]] = entry[1]
+	}
+
+	return envMap
 }
 
 func (db *DB) printVerbose(result sql.Result) {
@@ -491,7 +531,13 @@ func (db *DB) Rollback() error {
 
 	execMigration := func(tx dbutil.Transaction) error {
 		// rollback migration
-		result, err := tx.Exec(parsed.Down)
+		downScript, err := db.resolveRefs(parsed.Down, parsed.DownOptions.EnvVars())
+		if err != nil {
+			return err
+		}
+
+		result, err := tx.Exec(downScript)
+
 		if err != nil {
 			return err
 		} else if db.Verbose {
--- a/pkg/dbmate/db_test.go
+++ b/pkg/dbmate/db_test.go
@@ -237,7 +237,7 @@ func TestWaitBeforeVerbose(t *testing.T) {
 		`Applying: 20151129054053_test_migration.sql
 Rows affected: 1
 Applying: 20200227231541_test_posts.sql
-Rows affected: 0`)
+Rows affected: 1`)
 	require.Contains(t, output,
 		`Rolling back: 20200227231541_test_posts.sql
 Rows affected: 0`)
@@ -315,6 +315,11 @@ func TestUp(t *testing.T) {
 			err = sqlDB.QueryRow("select count(*) from users").Scan(&count)
 			require.NoError(t, err)
 			require.Equal(t, 1, count)
+
+			var fromEnvVar string
+			err = sqlDB.QueryRow("select name from posts where id=1").Scan(&fromEnvVar)
+			require.NoError(t, err)
+			require.Equal(t, "Yabba dabba doo!", fromEnvVar)
 		})
 	}
 }
--- a/pkg/dbmate/migration.go
+++ b/pkg/dbmate/migration.go
@@ -48,6 +48,7 @@ type ParsedMigration struct {
 // ParsedMigrationOptions is an interface for accessing migration options
 type ParsedMigrationOptions interface {
 	Transaction() bool
+	EnvVars() []string
 }
 
 type migrationOptions map[string]string
@@ -56,6 +57,21 @@ type migrationOptions map[string]string
 // Defaults to true.
 func (m migrationOptions) Transaction() bool {
 	return m["transaction"] != "false"
+}
+
+// EnvVars returns the list of env vars enabled to templating for this migration
+// Defaults to empty list.
+func (m migrationOptions) EnvVars() []string {
+	result := make([]string, 0)
+	entry := m["env-var"]
+
+	if entry != "" {
+		// decode CSV encoded var names
+		varNames := strings.Split(entry, ",")
+		// add to the slice
+		result = append(result, varNames...)
+	}
+	return result
 }
 
 var (
@@ -143,7 +159,18 @@ func parseMigrationOptions(contents string) ParsedMigrationOptions {
 
 		// if the syntax is well-formed, then store the key and value pair in options
 		if len(pair) == 2 {
-			options[pair[0]] = pair[1]
+			optKey := pair[0]
+			optValue := pair[1]
+			entry := options[optKey]
+			if entry != "" { // "env-var" entry already used
+				varNames := strings.Split(entry, ",")
+				// add new element to the slice
+				varNames = append(varNames, optValue)
+				// keep collected values
+				options[optKey] = strings.Join(varNames, ",")
+			} else { // first "env-var" entry
+				options[optKey] = optValue
+			}
 		}
 	}
 
--- a/pkg/dbmate/migration_test.go
+++ b/pkg/dbmate/migration_test.go
@@ -65,9 +65,11 @@ drop table users;
 
 		require.Equal(t, "--migrate:up\ncreate table users (id serial, name text);\n\n", parsed.Up)
 		require.Equal(t, true, parsed.UpOptions.Transaction())
+		require.Equal(t, []string{}, parsed.UpOptions.EnvVars())
 
 		require.Equal(t, "--migrate:down\ndrop table users;\n", parsed.Down)
 		require.Equal(t, true, parsed.DownOptions.Transaction())
+		require.Equal(t, []string{}, parsed.UpOptions.EnvVars())
 	})
 
 	t.Run("require up before down", func(t *testing.T) {
@@ -98,6 +100,40 @@ ALTER TYPE colors ADD VALUE 'orange' AFTER 'red';
 
 		require.Equal(t, "-- migrate:down transaction:false\nALTER TYPE colors ADD VALUE 'orange' AFTER 'red';\n", parsed.Down)
 		require.Equal(t, false, parsed.DownOptions.Transaction())
+	})
+
+	t.Run("support activating env vars", func(t *testing.T) {
+		migration := `-- migrate:up env-var:THE_ROLE env-var:THE_PASSWORD
+create role {{ .THE_ROLE }} login password {{ .THE_PASSWORD }};
+-- migrate:down env-var:THE_ROLE
+drop role {{ .THE_ROLE }};
+`
+
+		parsed, err := parseMigrationContents(migration)
+		require.Nil(t, err)
+
+		require.Equal(t, "-- migrate:up env-var:THE_ROLE env-var:THE_PASSWORD\ncreate role {{ .THE_ROLE }} login password {{ .THE_PASSWORD }};\n", parsed.Up)
+		require.Equal(t, []string{"THE_ROLE", "THE_PASSWORD"}, parsed.UpOptions.EnvVars())
+
+		require.Equal(t, "-- migrate:down env-var:THE_ROLE\ndrop role {{ .THE_ROLE }};\n", parsed.Down)
+		require.Equal(t, []string{"THE_ROLE"}, parsed.DownOptions.EnvVars())
+	})
+
+	t.Run("support activating env vars", func(t *testing.T) {
+		migration := `-- migrate:up env-var:THE_ROLE env-var:THE_PASSWORD
+create role {{ .THE_ROLE }} login password {{ .THE_PASSWORD }};
+-- migrate:down env-var:THE_ROLE
+drop role {{ .THE_ROLE }};
+`
+
+		parsed, err := parseMigrationContents(migration)
+		require.Nil(t, err)
+
+		require.Equal(t, "-- migrate:up env-var:THE_ROLE env-var:THE_PASSWORD\ncreate role {{ .THE_ROLE }} login password {{ .THE_PASSWORD }};\n", parsed.Up)
+		require.Equal(t, []string{"THE_ROLE", "THE_PASSWORD"}, parsed.UpOptions.EnvVars())
+
+		require.Equal(t, "-- migrate:down env-var:THE_ROLE\ndrop role {{ .THE_ROLE }};\n", parsed.Down)
+		require.Equal(t, []string{"THE_ROLE"}, parsed.DownOptions.EnvVars())
 	})
 
 	t.Run("require migrate blocks", func(t *testing.T) {
--- a/testdata/db/migrations/20200227231541_test_posts.sql
+++ b/testdata/db/migrations/20200227231541_test_posts.sql
@@ -1,8 +1,9 @@
--- migrate:up
+-- migrate:up env-var:YABBA_DABBA_DOO
 create table posts (
   id integer,
   name varchar(255)
 );
+insert into posts (id, name) values (1, '{{ .YABBA_DABBA_DOO }}');
 
 -- migrate:down
 drop table posts;


amacneil:
@dossy @gregwebs for review

dossy (changes requested):
I'd like to see a few additional test cases that will codify the following behaviors and help set expectations:

* What does dbmate do if the migration specifies and environment variable, and that variable is unset in the environment?
* What does dbmate do if a migration has a template string (e.g., `{{ .FOO }}`) and `env:FOO` is not specified in the migration?
* What does dbmate do if there are characters in the environment variable that should be escaped, e.g., `YABBA_DABBA_DOO="this ' is ' a string"`?  Is the expectation that the user executing dbmate _must_ ensure that all environment variable contents are properly escaped string values for the target database, and otherwise be susceptible to SQL injections and general errors due to improperly escaped strings?  Is this a reasonable expectation?  Will this behavior violate the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)? 

Also, please update [the README's section on "Migration files"](https://github.com/davidecavestro/dbmate/blob/env-vars/README.md#migration-files) documenting this new functionality.

davidecavestro:
> I'd like to see a few additional test cases that will codify the following behaviors and help set expectations:
> 
> * What does dbmate do if the migration specifies an environment variable (e.g., `env:FOO`), and that environment variable is unset in the environment at the time dbmate is executed?  What should the expected behavior/outcome be?  Silently interpolate an empty string?  Raise an error?  Something else entirely?
> * What does dbmate do if a migration has a template string (e.g., `{{ .FOO }}`) and `env:FOO` is not specified in the migration?  What should the expected behavior/outcome be?  Silently interpolate an empty string?  Raise an error?  Leave the template fragment as-is?  Something else entirely?
> * What does dbmate do if there are characters in the environment variable that should be escaped, e.g., `YABBA_DABBA_DOO="this ' is ' a string"`?  Is the expectation that the user executing dbmate _must_ ensure that all environment variable contents are properly escaped string values for the target database, and otherwise be susceptible to SQL injections and general errors due to improperly escaped strings?  Is this a reasonable expectation?  Will this behavior violate the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)?
> 
> Also, please update [the README's section on "Migration files"](https://github.com/davidecavestro/dbmate/blob/env-vars/README.md#migration-files) documenting this new functionality.

Good points: I'd like some more feedback before proceding with changes.

For the 1st and 2nd ones I propose to raise an error, i.e. we could simply add `.Option("missingkey=error")` to the template.
Migration authors can still set default values within the template expression in case they prefer, and we could document it. 

About SQL injections we could abandon templates in favor of query parameters/placeholders, passing env var values i.e. `tx.Exec(upScript, envVars)`.

IMHO the downsides of placeholders are:
- positional binding -> migration authors must pay attention to the relative position of env vars declarations
- the placeholder syntax varies between db drivers

AFAIK named parameters are only available throuch additional packages like https://github.com/jmoiron/sqlx, so if the mandate is keeping thing simple, that's not a viable option.

OTOH if we keep both templates and placeholders - leaving the migration author choose between risks of SQL injection and portability/positional declarations - we end up violating both the POLA and KISS principle.

gregwebs:
I agree that erroring on unset/unspecified is best.

Perhaps we should think about placeholders more? Then we don't need our own variable feature but instead a feature for working with placeholders.

gregwebs:
What's the actual downside of using sqlx?

gregwebs:
It seems placeholders may end up being too limited and templates will be needed.
The pgx author himself [created a migration system](https://github.com/jackc/tern) and it has templates.

gregwebs:
The migration system should be referencing static values in its template variables rather than user input, so SQL injection would have to be an insider attack that uses a static value that makes it through code review (or an environment variable and a controlled deployed environment)- quite different from what we [normally think of for SQL injection](https://xkcd.com/327/).

With respect to escaping, I think we can put the burden on the user that wants to use the template system. Over time we can provide some helper functions for different types of escaping.

I think this PR is good once errors are raised for missing variables.

davidecavestro:
Is it ok if I move the env var refs resolution logic into an internal package, in order to unit test it but still keeping private access.
image

gregwebs:
Few things I like hearing more than talk of testing. I also like smaller packages. It's possible other maintainers don't want to introduce a new pattern and can speak up, but I am fine with it.

davidecavestro:
Ouch, I forgot to update docs, but I'll do it soon

davidecavestro:
So I added some references to the `env` option within the _Migration Options_ section of the readme.

I also played with naive sql injection attempts and saw they could be mitigated using the `js` function, hence I mentioned it in the docs and added some explicit tests.

pkg/dbmate/internal/envvars_test.go
```diff
@@ -71,6 +88,25 @@ func TestResolveWithSqlInjection(t *testing.T) {
 	resolved, err := internal.ResolveRefs(parseUp, parsedUpOptsEnvVars, envMap)
 
 	require.NoError(t, err)
-	// sql injectio in not prevented here
+	// sql injectio is not prevented here
```
dossy, line 91: fix: `injectio` -> `injection`

README.md
```diff
@@ -323,8 +323,40 @@ Writing: ./db/schema.sql
 
 dbmate supports options passed to a migration block in the form of `key:value` pairs. List of supported options:
 
+- `env`
 - `transaction`
 
+**env**
+
+`env` is useful if you need to inject some environment variable into the SQL. Only environment variables explicitly listed will be available for templating:
```
dossy, line 331: fix: `environment variable into` -> `environment variables into`

README.md
```diff
@@ -323,8 +323,40 @@ Writing: ./db/schema.sql
...
+
+```sql
+-- migrate:up env:ADMIN_PASSWORD
+create role 'adminuser' login password '{{ .ADMIN_PASSWORD }}';
+```
+
+`env` supports multiple occurrencies. If a referenced variable is not available the command will return an error:
```
dossy, line 338: fix: `occurrencies` -> `occurrences`

dossy (changes requested):
Looking good.  Some minor English language corrections requested, but otherwise 👍.

dossy (approved):

README.md
```diff
@@ -323,8 +323,40 @@ Writing: ./db/schema.sql
...
+```
+
+Variables injection leverages Go's [template](https://pkg.go.dev/text/template) features. For example you can provide default values in order to prevent errors:
+
+```sql
+-- migrate:up env:THE_ROLE env:THE_PASSWORD
+create role '{{ or (index . "THE_ROLE") "adminuser" }}' login password '{{ .THE_PASSWORD }}';
+```
+
+The `js` function may help to mitigate risks of SQL injection:
```
gregwebs, line 353: sorry if I missed the discussion- how does this mitigate SQL injections risks? Can't I insert a string with a single quote still?

dossy: The [go/template functions](https://pkg.go.dev/text/template#hdr-Functions) includes `js` which is described as:

```
js
	Returns the escaped JavaScript equivalent of the textual
	representation of its arguments.
```

While this may not necessarily be sufficient to prevent *all* possible SQL injection attacks (see: Unicode homoglyph injection attacks aka Unicode smuggling [1](https://hackernoon.com/%CA%BC-%C5%9B%E2%84%87%E2%84%92%E2%84%87%E2%84%82%CA%88-how-unicode-homoglyphs-will-break-your-custom-sql-injection-sanitizing-functions-1224377f7b51) [2](https://www.dionach.com/en-us/blog/fun-with-sql-injection-using-unicode-smuggling/), character set transcoding SQL injection attacks [3](https://shiflett.org/blog/2006/addslashes-versus-mysql-real-escape-string)), it prevents the most trivial kind where an unescaped single quote `'` is in a string that will be interpolated into the SQL query, by escaping it with a backslash, thus turning `'` into `\'`.

davidecavestro: The trivial case of unescaped single quote is covered by [TestResolveMitigatingSqlInjection](https://github.com/amacneil/dbmate/pull/439/files#diff-e38ec788aaa9a3ec91568145000f7d7666bd296e62d65dadd1b32acc63bf9652R95).

gregwebs: nice test case!

dossy:
Before merging this, I'm only very slightly concerned that we are technically introducing a backwards-incompatible change, in that if anyone actually has `{{` and `}}` strings in their migrations for some reason, that would previously be interpreted literally as plain text, but now will be parsed and processed with `text/template` and will likely cause errors.

I suspect the likelihood of anyone having `{{` and `}}` in their migration files is unlikely, but it may not be zero probability.

@amacneil, how have you handled introducing backwards-incompatible changes into releases?  Does the dbmate project follow [semantic versioning](https://semver.org/) for its version numbering?  If yes, then this probably means needing to bump the version to v3.0.0, and not v2.4.0, which seems a bit much considering the amount of functionality introduced by this change.

Perhaps we should put migration file parsing behind a command line argument, default off, which would allow this change to be fully backwards-compatible?  @davidecavestro, how do you feel about that?  I propose `--use-env` or `--enable-templates` as the switch to control this.

davidecavestro:
> Before merging this, I'm only very slightly concerned that we are technically introducing a backwards-incompatible change, in that if anyone actually has `{{` and `}}` strings in their migrations for some reason, that would previously be interpreted literally as plain text, but now will be parsed and processed with `text/template` and will likely cause errors.
> 
> I suspect the likelihood of anyone having `{{` and `}}` in their migration files already is unlikely, but it may not be zero probability.
> 
> @amacneil, how have you handled introducing backwards-incompatible changes into releases? Does the dbmate project follow [semantic versioning](https://semver.org/) for its version numbering? If yes, then this probably means needing to bump the version to v3.0.0, and not v2.4.0, which seems a bit much considering the amount of functionality introduced by this change.
> 
> Perhaps we should put migration file parsing behind a command line argument, default off, which would allow this change to be fully backwards-compatible? @davidecavestro & @gregwebs, how do you feel about that? I propose `--use-env` or `--enable-templates` as the switch to control this.

We could also enable templating just when *at least* one occurrence of _env:VAR_NAME_ has been set into the header

gregwebs:
> We could also enable templating just when at least one occurrence of env:VAR_NAME has been set into the header.

That seems like a great way to do it- it ends up being enabled automatically when needed on a per-migration basis- so it would be fully backwards compatible.

dossy:
I'm indifferent to the two options.  My concern is guaranteeing backwards-compatibility in all cases, which both options, (a) only enable template parsing when a migration defines an `env` variable for interpolation, or (b) only enable template parsing when a command line argument is specified, satisfy.

I _would_ vote in favor of enabling Go template parsing with `--enable-templating` over auto-enabling it only when an `env` variable is specified, because there are other potentially useful ways that writing migration scripts using Go templating aside from interpolating environment variables, so it would be nice to have a way to enable it without having to define a placeholder env var in the migration script.

While I personally dislike migration scripts that are not plain text with literal values because it potentially introduces issues with repeatability (a future execution of the unmodified migration script can result in a different migration), I understand others do not try to abide by this constraint and a template-driven migration script might be really useful.

Introducing Go template parsing in general opens the door for a wide range of functionality we could make available to migration script writers aside from only interpolating environment variables.

We could even go one step further and remove the entire `env:var` mechanism per-migration but keep the `--enable-templates` functionality, and instead expose [`os.Getenv`](https://pkg.go.dev/os#Getenv) and [`os.ExpandEnv`](https://pkg.go.dev/os#ExpandEnv) to `text/template` using [template.Funcs](https://pkg.go.dev/text/template#Template.Funcs) and a [template.FuncMap](https://pkg.go.dev/text/template#FuncMap), like [Hugo](https://gohugo.io/functions/getenv/) or [gomplate](https://docs.gomplate.ca/functions/env/) does.

Users could then use `{{ os.Getenv "ENV_VAR_NAME" "defaultValue" }}` in their migration script.  This would mean that migration script writers would have unrestricted to any environment variable, which may be an issue.  Hugo implements a [Security Policy](https://gohugo.io/about/security-model/#security-policy) that has a default filter on what environment variables can be accessed by default.  If necessary, we could do something similar, and the filter could have a reasonable default value and overridable with a command-line argument like `--os-env-regexp` or something.

Something to consider when deciding between enabling templating if and only if  `env:var` is specified on a migration, or when `--enable-templates` is specified as a command-line argument.

gregwebs:
If you want reproducibility, the current system has it- it just needs to record the values of the variables used.

We can do auto-enabling now and then still add the flag in the future if we support reading in template values that are not pre-declared.

amacneil:
> @amacneil, how have you handled introducing backwards-incompatible changes into releases? Does the dbmate project follow [semantic versioning](https://semver.org/) for its version numbering?

Yes, following semver. In general I try not to make backwards incompatible changes to the CLI/migration spec.

> If yes, then this probably means needing to bump the version to v3.0.0, and not v2.4.0, which seems a bit much considering the amount of functionality introduced by this change.

Agreed, better if we can do it in a backwards compatible way.

> Introducing Go template parsing in general opens the door for a wide range of functionality we could make available to migration script writers aside from only interpolating environment variables.
> 
> We could even go one step further and remove the entire env:var mechanism per-migration but keep the --enable-templates functionality, and instead expose os.Getenv and os.ExpandEnv to text/template using template.Funcs and a template.FuncMap, like Hugo or gomplate does.

My initial reaction is that this goes against dbmate's philosophy of keeping things simple. Now we are introducing a whole templating language (and one that will be completely unfamiliar to non-Go developers). I've [stated](https://github.com/amacneil/dbmate/issues/118#issuecomment-699521772) [in](https://github.com/amacneil/dbmate/issues/118#issuecomment-699577136) the [past](https://github.com/amacneil/dbmate/issues/118#issuecomment-827191112) that I think env vars in migrations are not a great idea, but if we are going to support it, then I think we should strictly limit this to env vars and not support any other functions. 

> We could also enable templating just when at least one occurrence of env:VAR_NAME has been set into the header.

Agreed, this seems sufficient to ensure backwards compatibility (existing migrations will not break, future migrations that use this feature will need to be aware of it.

> I also played with naive sql injection attempts and saw they could be mitigated using the js function, hence I mentioned it in the docs and added some explicit tests.

I disagree with adding the `js` function, because again it is something new for the user to learn, and javascript escaping is not sql escaping so it won't work in all cases. Doing a half assed job of sql escaping is worse than not doing it at all. And we cannot easily provide a sql escaping function because it is different for each database and depending on what type of escaping you need (string vs identifier, etc).

Here is an alternative solution to address the problem of sql injection:

```
-- migrate:up unsafe-env:ROLE env:PASSWORD
CREATE USER {{.ROLE}} WITH PASSWORD {{.PASSWORD}};
```

- Keep things simple: no fancy functions for the user to learn, only variables are available
- `env:FOO` passes the variable through database-specific string escaping
- `unsafe-env:FOO` means I know what I am doing and disable string escaping for that variable (e.g. you need it to create an identifier in postgres). Naming convention is inspired by web standards, e.g. CSP's [unsafe-eval](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_eval_expressions) and React's [dangerouslySetInnerHTML](https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html)
- if you specify the same variable twice or otherwise provide an invalid `--migrate` directive we should abort with an error

amacneil:
> The migration system should be referencing static values in its template variables rather than user input, so SQL injection would have to be an insider attack that uses a static value that makes it through code review (or an environment variable and a controlled deployed environment)- quite different from what we normally think of for SQL injection.
> 
> With respect to escaping, I think we can put the burden on the user that wants to use the template system. Over time we can provide some helper functions for different types of escaping.

I do agree with this, and that would be a perfectly reasonable stance. However, the only real use case I have heard articulated for this feature seems to revolve around not storing passwords in your migration files and injecting them at runtime. In that case, it would be nice if your long secure passwords don't accidentally trip up migrations due to accidental sql injection (even if it will not be exploited). So I think doing my automatic string escaping proposal above would be worthwhile, as long as we give people an escape hatch.

README.md
amacneil: I added a comment in the main PR thread - I'm not a fan of introducing functions to the template syntax, or encouraging use of js escaping functions for sql escaping.

gregwebs:
Another use case that I saw was to use a different schema name for a different environment. So rather than inserting a string, it would be inserting a database name.

gregwebs:
We could expose a new command "dbmate template" for those that want to do ad hoc templating. It would be used as a pre phase- take a file as an input and output, a file.

Then, for this environment, variable approach we could only support placeholders values such as strings.

davidecavestro:
ok guys, it's plenty of viable options here, and it's certainly up to you maintainers driving the change with proper vision.

I drafted this PR with the aim to keep things really simple, avoiding to tackle with SQL injection for many reasons.
Then during reviews I added an example of standard `js` function usage along with a specific test case.
I can easily remove them to avoid confusion.

That said, I totally understand if you prefer to postpone and or/restart designing the whole thing from scratch to get better features.

In other words I'm not in a hurry to have this feature: I contributed this PR just to give back and keep hands on golang.
Just my two cents.

gregwebs:
Thanks for the understanding. It is obvious we are lacking in a process to agree on how new features should behave before they start being developed.

I guess one question that is being asked now is- what does the implementation look like to focus on the password insert case and use a placeholder for a string? Are there inherit difficulties or is there a straight-forward approach.

As stated, I think even if we go that route we can still expose full templating for those that want it.

gregwebs:
I don't think there's anything inherently unsafe here- again this isn't dealing with arbitrary user input- it's variables values are known ahead of time, except for passwords. Randomly generated passwords aren't really a threat- this would have to be a user supplied password. That just needs to be given as a string. It might make more sense to use terms like escaped, quoted, raw, text, etc and make a note in the docs about passwords.

dossy:
> I don't think there's anything inherently unsafe here [...]

In this context, "unsafe" communicates that you, the user, are wholly and entirely responsible for escaping the input string as necessary, that passing through values **should be considered an unsafe operation** if you would have otherwise expected your strings to be escaped and quoted for you.  Essentially, "unsafe" is a warning signal to the operator that there are NO protections offered here: you are wielding a footgun.  You are responsible for how you aim it.

amacneil:
If we can't disable the Go template helpers, then how about we just do our own templating using `${}` syntax? It doesn't need to be complicated.

E.g. this would work fine:

```sql
-- migrate:up unsafe-env:ROLE env:PASSWORD
CREATE USER ${ROLE} WITH PASSWORD ${PASSWORD};
```

While we're here, it should probably be an error to run a migration that specifies an environment variable if that environment variable is unset (empty is ok, unset should be an error).

DevopsMercenary:
I found a nuance / issue that I wanted to let you know about.

In a db migration, if I want to use variables I follow this syntax....

```
-- migrate:up env:DBNAME env:OWNER
```

Therefore I would assume that for the down the syntax would be the same

```
-- migrate:down env:DBNAME env:OWNER
```

But I get an error...

```
Error: dbmate requires each migration to define a down block with '-- migrate:down'
```

BUT I did figure out that the following syntax works...

```
-- migrate:down env:DBNAME,OWNER
```
If you need any more details please let me know.  I have the latest from your branch and I have a current version of GoLang

[greg@CM-LT-0007 dbmate (env-vars)]$ git status
On branch env-vars
Your branch is up to date with 'origin/env-vars'.
nothing to commit, working tree clean

go version go1.20.6 darwin/arm64

o-lee:
Will this PR be released soon? Thanks.

dossy:
> Will this PR be released soon? Thanks.

@o-lee There's no expected release date for this PR.

At a minimum, I think [this](https://github.com/amacneil/dbmate/pull/439#issuecomment-1558586436) needs to be implemented before this is safe to merge:

> > We could also enable templating just when at least one occurrence of env:VAR_NAME has been set into the header.
> 
> Agreed, this seems sufficient to ensure backwards compatibility (existing migrations will not break, future migrations that use this feature will need to be aware of it.

Or, alternatively, [this](https://github.com/amacneil/dbmate/pull/439#issuecomment-1557291240):

> Perhaps we should put migration file parsing behind a command line argument, default off, which would allow this change to be fully backwards-compatible? @davidecavestro & @gregwebs, how do you feel about that? I propose `--use-env` or `--enable-templates` as the switch to control this.

And, someone needs to investigate @DevopsMercenary's [issue](https://github.com/amacneil/dbmate/pull/439#issuecomment-1662899954) and determine what's going on.

We need someone to work on these.

dossy (changes requested):
Based on @DevopsMercenary's [comment](https://github.com/amacneil/dbmate/pull/439#issuecomment-1662899954), it's clear that we need more test cases.  Our current tests only cover `migrate:down` with [only one variable](https://github.com/amacneil/dbmate/pull/439/files#diff-b72af22c75eff5a7a050904af16e75a69aacdc9481296fdbf834335fd3589963R121-R122).  At a minimum, we need a test with one, and a test with more than one.

o-lee:
@dossy, thanks for the update.

plentydone:
Would there be appetite for me to pick this back up and address all of [these](https://github.com/amacneil/dbmate/pull/439#issuecomment-1896155283) comments?

dossy:
> Would there be appetite for me to pick this back up and address all of [these](https://github.com/amacneil/dbmate/pull/439#issuecomment-1896155283) comments?

As I have no horse in this race—I have no need for this functionality, nor plan to use it—I am trying to limit my feedback to technical issues and concerns.

Whether there's demand for this by others who are waiting to use it, they should speak up and represent themselves.

davidecavestro:
> Would there be appetite for me to pick this back up and address all of [these](https://github.com/amacneil/dbmate/pull/439#issuecomment-1896155283) comments?

I gave up a long time ago, so I'm very happy if you're willing to squeeze value out of it.

DevopsMercenary:
I'm using this branch and would still really like to see this feature.  It really helps with keeping the scripts DRY ( don't repeat yourself ) an important principal we prefer to follow.

amacneil:
I don't have a personal need for this feature, but happy for someone to take it across the finish line (including tests etc).

I stand by the syntax recommendation I made in this comment https://github.com/amacneil/dbmate/pull/439#issuecomment-1590117696

```sql
-- migrate:up unsafe-env:ROLE env:PASSWORD
CREATE USER ${ROLE} WITH PASSWORD ${PASSWORD};
```

- no dependency on golang templates strings (unfamiliar syntax to many users, `${}` feels much simpler and more language-agnostic)
- env vars must be explicitly imported at top of the migration
- `env:FOO` imports string-escaped value
- `unsafe-env:FOO` imports unescaped value

plentydone:
@amacneil  coincidentally I started a branch last night where I replaced templating with `os.Expand` . I was only working on the unsafe version but I can look at adding engine-appropriate quoting.

plentydone:
I dropped this a while back after considering complexity and feature surface area it added to what is a rather streamlined codebase. The next time I use this tool I'll use envsubst beforehand if I need it.

Issue #159, opened 2020-09-24
Document migration order behaviour.


kevincox:
I can see that the migration ID is the current date by default which leads me to believe that migrations are applied in order. However there is a lack of documentation of what happens if an out-of-order migration is identified.

For example image that my DB is at version 1. Then two developers make feature branches and add migrations 2 and 3. If 3 merges and migrates first what happens when migration 2 gets merged?

It seems that there are 2 major options here:
1. Reject migration 2 ("Error: Out of order migration 2 for DB version 3.")
2. Apply migration 2.

I think option 2 is generally preferable however it does cause complications if the two migrations conflict. If you are lucky you will get an error on conflict, however the rare case could be that you get a different result applying 3 then 2 as opposed to 2 then 3. This would lead in the production DB having a different schema than developer machines which could be a source of bugs.

1 is annoying upfront but can be fixed by manually renaming migration 2 to come after migration 3. However depending on the rate of change of your project this Compare and Set style work can quickly become tedious.

Either way the current behaviour should be documented.

amacneil:
Correct - dbmate's behavior is that we will apply all unapplied migrations in numeric order (your option 2). We intentionally do not error out here, because the situation you describe is extremely common (developers working on separate PRs which land out of order), and 9 times out of 10 applying them in a different order is a non-issue. Also, in the unlikely event that there is a conflict, it will usually result in an error rather than an out of sync production db (e.g. if two developers try to rename the same column).

In my experience this works fine for teams up to a few dozen developers working on a single project, and beyond that you probably want a dedicated release engineer/manager anyway.

Thanks for the comment, I will document this behavior.

kevincox:
Thanks, the described behaviour is what I would want for my use cases. Just was missing that detail when evaluating the project.

amacneil:
Great!

I'll leave a note here for future readers that if anyone really does desire the first option (reject out of order migrations), it would be easy enough to add a `--strict-order` flag or something. Happy to accept a PR.

PR #441, opened 2023-05-04
Add --strict flag

philip-hartmann:
This adds the `--strict-order` flag which ignores all out of order pending migrations. With the flag set `dbmate up` and `dbmate migrate` only apply pending migrations which succeed the latest applied migration - in other words, only pending migrations with a higher version number than the latest applied migration will be used. `dbmate --strict-order status` also only lists in order pending migrations. Rollbacks work as usual.

Refs: #159

### DIFF
--- a/main.go
+++ b/main.go
@@ -102,8 +102,8 @@ func NewApp() *cli.App {
 			Usage: "Create database (if necessary) and migrate to the latest version",
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
-					Name:    "strict-order",
-					EnvVars: []string{"DBMATE_STRICT_ORDER"},
+					Name:    "strict",
+					EnvVars: []string{"DBMATE_STRICT"},
 					Usage:   "ignore out of order pending migrations",
 				},
 				&cli.BoolFlag{
@@ -114,7 +114,7 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				db.StrictOrder = c.Bool("strict-order")
+				db.Strict = c.Bool("strict")
 				db.Verbose = c.Bool("verbose")
 				return db.CreateAndMigrate()
 			}),
@@ -138,8 +138,8 @@ func NewApp() *cli.App {
 			Usage: "Migrate to the latest version",
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
-					Name:    "strict-order",
-					EnvVars: []string{"DBMATE_STRICT_ORDER"},
+					Name:    "strict",
+					EnvVars: []string{"DBMATE_STRICT"},
 					Usage:   "ignore out of order pending migrations",
 				},
 				&cli.BoolFlag{
@@ -150,7 +150,7 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				db.StrictOrder = c.Bool("strict-order")
+				db.Strict = c.Bool("strict")
 				db.Verbose = c.Bool("verbose")
 				return db.Migrate()
 			}),
@@ -177,8 +177,8 @@ func NewApp() *cli.App {
 			Usage: "List applied and pending migrations",
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
-					Name:    "strict-order",
-					EnvVars: []string{"DBMATE_STRICT_ORDER"},
+					Name:    "strict",
+					EnvVars: []string{"DBMATE_STRICT"},
 					Usage:   "ignore out of order pending migrations",
 				},
 				&cli.BoolFlag{
@@ -191,7 +191,7 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
-				db.StrictOrder = c.Bool("strict-order")
+				db.Strict = c.Bool("strict")
 				setExitCode := c.Bool("exit-code")
 				quiet := c.Bool("quiet")
 				if quiet {
--- a/main.go
+++ b/main.go
@@ -62,11 +62,6 @@ func NewApp() *cli.App {
 			Value:   defaultDB.MigrationsTableName,
 			Usage:   "specify the database table to record migrations in",
 		},
-		&cli.BoolFlag{
-			Name:    "strict-order",
-			EnvVars: []string{"DBMATE_STRICT_ORDER"},
-			Usage:   "ignore out of order pending migrations",
-		},
 		&cli.StringFlag{
 			Name:    "schema-file",
 			Aliases: []string{"s"},
@@ -107,6 +102,11 @@ func NewApp() *cli.App {
 			Usage: "Create database (if necessary) and migrate to the latest version",
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
+					Name:    "strict-order",
+					EnvVars: []string{"DBMATE_STRICT_ORDER"},
+					Usage:   "ignore out of order pending migrations",
+				},
+				&cli.BoolFlag{
 					Name:    "verbose",
 					Aliases: []string{"v"},
 					EnvVars: []string{"DBMATE_VERBOSE"},
@@ -114,6 +114,7 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
+				db.StrictOrder = c.Bool("strict-order")
 				db.Verbose = c.Bool("verbose")
 				return db.CreateAndMigrate()
 			}),
@@ -137,6 +138,11 @@ func NewApp() *cli.App {
 			Usage: "Migrate to the latest version",
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
+					Name:    "strict-order",
+					EnvVars: []string{"DBMATE_STRICT_ORDER"},
+					Usage:   "ignore out of order pending migrations",
+				},
+				&cli.BoolFlag{
 					Name:    "verbose",
 					Aliases: []string{"v"},
 					EnvVars: []string{"DBMATE_VERBOSE"},
@@ -144,6 +150,7 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
+				db.StrictOrder = c.Bool("strict-order")
 				db.Verbose = c.Bool("verbose")
 				return db.Migrate()
 			}),
@@ -170,6 +177,11 @@ func NewApp() *cli.App {
 			Usage: "List applied and pending migrations",
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
+					Name:    "strict-order",
+					EnvVars: []string{"DBMATE_STRICT_ORDER"},
+					Usage:   "ignore out of order pending migrations",
+				},
+				&cli.BoolFlag{
 					Name:  "exit-code",
 					Usage: "return 1 if there are pending migrations",
 				},
@@ -179,6 +191,7 @@ func NewApp() *cli.App {
 				},
 			},
 			Action: action(func(db *dbmate.DB, c *cli.Context) error {
+				db.StrictOrder = c.Bool("strict-order")
 				setExitCode := c.Bool("exit-code")
 				quiet := c.Bool("quiet")
 				if quiet {
@@ -238,7 +251,6 @@ func action(f func(*dbmate.DB, *cli.Context) error) cli.ActionFunc {
 		db.AutoDumpSchema = !c.Bool("no-dump-schema")
 		db.MigrationsDir = c.StringSlice("migrations-dir")
 		db.MigrationsTableName = c.String("migrations-table")
-		db.StrictOrder = c.Bool("strict-order")
 		db.SchemaFile = c.String("schema-file")
 		db.WaitBefore = c.Bool("wait")
 		waitTimeout := c.Duration("wait-timeout")
--- a/main.go
+++ b/main.go
@@ -104,7 +104,7 @@ func NewApp() *cli.App {
 				&cli.BoolFlag{
 					Name:    "strict",
 					EnvVars: []string{"DBMATE_STRICT"},
-					Usage:   "apply migrations strictly in numerical order",
+					Usage:   "fail if migrations would be applied out of order",
 				},
 				&cli.BoolFlag{
 					Name:    "verbose",
@@ -140,7 +140,7 @@ func NewApp() *cli.App {
 				&cli.BoolFlag{
 					Name:    "strict",
 					EnvVars: []string{"DBMATE_STRICT"},
-					Usage:   "apply migrations strictly in numerical order",
+					Usage:   "fail if migrations would be applied out of order",
 				},
 				&cli.BoolFlag{
 					Name:    "verbose",
--- a/main.go
+++ b/main.go
@@ -62,6 +62,11 @@ func NewApp() *cli.App {
 			Value:   defaultDB.MigrationsTableName,
 			Usage:   "specify the database table to record migrations in",
 		},
+		&cli.BoolFlag{
+			Name:    "strict-order",
+			EnvVars: []string{"DBMATE_STRICT_ORDER"},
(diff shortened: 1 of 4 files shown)


gregwebs:
This code change looks good to me. From a UX perspective I find `--strict-order` a confusing name. Something like `--new-only` would make more sense to me.

It's also wort noting that my PR #438 would allow you to select a range of migrations, such as the range that doesn't include some older migrations- so a slightly more manual version of this. For me it seems like it would work because this use case wouldn't come up a lot.

philip-hartmann:
@amacneil proposed the name `--strict-order` for the flag in #159, so I went with it. The flag comes in handy as a permanent flag you would choose to make your migration strategy more _strict_, to avoid migration conflicts mentioned in #159 and kevlarr/jrny/issues/17.

I also do not see `--strict-order` in conflict with #438, as they both tackle different use cases.

amacneil:
My linked comment was 3 years ago, but I think probably just `--strict` is even better for the flag name? It seems to be in the spirit of "be strict about applying migrations" in order, and fail if you can't".

I think this should be a flag on the verb, rather than a global flag. It only applies to the `up`/`migrate` command.

I'll let others do code review @gregwebs @dossy

main.go
```diff
@@ -164,6 +176,11 @@ func NewApp() *cli.App {
 			Name:  "status",
 			Usage: "List applied and pending migrations",
 			Flags: []cli.Flag{
+				&cli.BoolFlag{
```
gregwebs, line 179: There is a precedent for duplication with the verbose flag, but it seems like it would be better to define the flag just once

```
strictFlag := cli.BoolFlag{...}
flags : []cli.Flag{&strictFlag}
```

gregwebs:
I think there would be some benefit to instead of just ignoring migrations to have another state in the UI to show that migrations were cutoff due to the strict setting. Users may set this as an env variable and then get confused as to why migrations aren't showing up.

That being said, users that set this will probably be more advanced and I don't have an answer as to what exactly to show users (just that there is a strict cutoff, or a different kind of marking instead of applied/unapplied).
As is this makes for a simple implementation and UI improvements could still be made with future changes.

dossy (approved):
Code looks good to me.

For the record, I would never expect something called "strict" to ever ignore anything.  "--ignore-outdated" may be more verbose but I would feel more confident guessing exactly what the behavior is, vs. "strict" which I would expect would result in *a hard error* if there were unapplied migrations or anything else that would not pass a "strict" test.

But, if the consensus is that "strict" is unambiguous and the implemented behavior matches the expectation of others aside from me, then I'm okay with it.

pkg/dbmate/db.go
```diff
@@ -412,6 +415,13 @@ func (db *DB) FindMigrations() ([]Migration, error) {
 		}
 	}
 
+	var latestAppliedMigration string
+	for migrationVersion := range appliedMigrations {
+		if migrationVersion > latestAppliedMigration {
+			latestAppliedMigration = migrationVersion
+		}
```
dossy, line 422: I will say that I think some automated tests to clarify how this change will behave could be useful to clarify expectations.

Is Go's string comparison implementation robust enough that `"100_foo" > "2_bar"` will be true?  [Looks like it isn't.](https://go.dev/play/p/0Fh_jGFUTcI)  What would you expect dbmate to do in this situation?  In this particular situation, the code above would consider "2_bar" the latest applied migration, even if "100_bar" had also been applied.  Is that what you would expect?

philip-hartmann: The version comparison seems to be in line with how dbmate already operates. `FindMigrations()` returns a sorted slice in the end but also sorts the slice similarly just by file name. Dbmate already creates its migrations with a sortable file name, such as `20151127184807_create_users_table.sql`. So as long as all migrations are created by dbmate that will not be a problem. I agree with you that the behaviour should be properly described. The order of execution for pending migrations is described as _in numerical order_ in the [running migrations](https://github.com/amacneil/dbmate#running-migrations) section; but it's in alphanumeric order? A proper description of the flag should also be added to that section. I will look into some possible additional test cases but it seems they would also need to test the general behaviour of dmate with migration files that do not use the same file naming scheme?

Regarding the expensive loop: The applied migrations are indeed queried in descending order in all used database drivers but they are stored in an unsorted `map[string]bool` afterwards. I do not see another way to get the latest applied migration without too many unrelated changes that could affect other PRs, except doing another database query which does not seem practical.

main.go
philip-hartmann: That could really tidy up the duplicate flag usage. I was already a little annoyed writing that part but I tried to do it similar to the verbose flag. I will change it 👍

philip-hartmann:
The more I think about the flag name the more I am bothered by it. @gregwebs and @dossy raise good points about the unambiguousness of the flag, both on the user level and in the code itself too. A more verbose name would definitely not hurt and would make a condition like `if db.Strict` much more expressive. Also, could there be a case that _strict_ will be useful for something else in the future or is _strict_ in the context of dbmate unambiguous with applying migrations in order?

pkg/dbmate/db.go
gregwebs: yeah, this looked suspicious to me as well but then I looked farther down and saw that dbmate is doing the same string comparison on filename for its sorting.

I also looked at this looping and yeah it is quite odd that the current interface returns a map of applied migrations instead of just a slice. I imagine at the time it seemed more efficient, but now it ends up being less efficient :)

dossy: > [...] The order of execution for pending migrations is described as in numerical order in the [running migrations](https://github.com/amacneil/dbmate#running-migrations) section; but it's in alphanumeric order? [...]

Yeah, the documentation is potentially misleading, as it's in lexicographical order, which only coincidentally is in natural numeric order if the number-portion of the version number is formatted with the same number of characters, which it will be if you use dbmate-generated migration filenames, but if you want to generate your own migration filenames, this might be a source of surprise when it's not performing an actual natural sort.

Anyway, this is a potential gotcha for anyone using dbmate, not necessarily specific to this PR, so I'm happy we had this discussion but nothing actionable needs to be done with respect to this PR.  Thanks for hearing me out.

> [...] Regarding the expensive loop: The applied migrations are indeed queried in descending order in all used database drivers but they are stored in an unsorted map[string]bool afterwards. [...]

You are right, my mistake.

It's unfortunate that Go doesn't ship with an ordered map implementation out of the box.

I suppose having `Driver.SelectMigrations()` return a `[]string` slice is unappealing since using `sort.Search()` to test for existence of elements in the slice is a bit clumsy compared to indexing into a map and checking the return value, and having an O(1) operation to check for existence is probably more important since that's probably the more frequently executed use case, anyway.

amacneil:
Actually, I didn't fully comprehend the PR description, but I was expecting this behavior too:

> "strict" which I would expect would result in a fatal error if there were unapplied migrations or anything else that would not pass a "strict" test.

Would that be a better solution, rather than simply ignoring out of order migrations? If any migrations have a lower version number than the current "latest" version, then throw an error and exit 1?

Combined with something like https://github.com/amacneil/dbmate/pull/434 the user could then resolve the issue by manually applying migrations.

esceer:
Is this strict feature planned to be merged anytime soon? We are eager to use it once available.

x80486:
I think it would be great if when this is merged, a `fix` command could be implemented so the migrations that are borked can be renamed correctly and "align" them automatically.

philip-hartmann:
> Would that be a better solution, rather than simply ignoring out of order migrations? If any migrations have a lower version number than the current "latest" version, then throw an error and exit 1?

Sounds reasonable. I've made the appropriate changes. Strict mode throws if a pending migration is not strictly in numerical order with already applied migrations.

Migrations with the same version number are not yet properly checked though. Dbmate only saves the version number of applied migrations and not the full filename. Therefore we cannot properly compare pending migrations with already applied migrations.

dossy:
> Migrations with the same version number are not yet properly checked though. Dbmate only saves the version number of applied migrations and not the full filename. Therefore we cannot properly compare pending migrations with already applied migrations.

In my opinion, version numbers should be unique, and two files with the same version number should raise an error and halt.

This change would break backwards-compatibility, unfortunately.

What do others think?

amacneil:
> In my opinion, version numbers should be unique, and two files with the same version number should raise an error and halt.
> This change would break backwards-compatibility, unfortunately.

I think two files with the same version number should be considered an error always. We could introduce it behind a different flag and enable it by default in the next major version bump. However, I think this is an unrelated problem to the current PR.

main.go
```diff
@@ -101,6 +101,11 @@ func NewApp() *cli.App {
 			Name:  "up",
 			Usage: "Create database (if necessary) and migrate to the latest version",
 			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:    "strict",
+					EnvVars: []string{"DBMATE_STRICT"},
+					Usage:   "apply migrations strictly in numerical order",
```
amacneil, line 107: I think we need a better description here. I put one suggestion below. The flag is not "apply migrations strictly in numerical order", because we always do that. The point is that we fail if existing applied migrations have been applied out of order, if I understand correctly?

```suggestion
					Usage:   "fail if existing migrations have been applied out of order",
```

pkg/dbmate/db.go
```diff
@@ -441,6 +453,10 @@ func (db *DB) FindMigrations() ([]Migration, error) {
 				migration.Applied = true
 			}
 
+			if db.Strict && !migration.Applied && migration.Version < latestAppliedMigration {
+				return nil, fmt.Errorf("cannot apply migration `%s` after `%s` in --strict mode, migrations have to be in numerical order", migration.Version, latestAppliedMigration)
```
amacneil, line 457: similar to my other comment, I think this error message could be a bit more helpful. we can mirror whatever words are selected for the flag description.

amacneil:
I'm 👍 on the current functionality of this PR, after we resolve my comment about the words used to describe the feature and error message.

main.go
philip-hartmann: True, the description should be more precise. The current PR only applies for new/pending migrations. It fails if a new migration has a smaller version number than the already applied migration with the highest version number.

pkg/dbmate/db.go
```diff
@@ -441,6 +453,10 @@ func (db *DB) FindMigrations() ([]Migration, error) {
 				migration.Applied = true
 			}
 
+			if db.Strict && !migration.Applied && migration.Version < latestAppliedMigration {
```
amacneil, line 456: This is not the right place for this check to live. `FindMigrations()` should not fail regardless of what migrations have or haven't been applied.

The check should be implemented in the `Migrate()` function

amacneil (changes requested):

pkg/dbmate/db.go
philip-hartmann: Without an additional db query the check needs access to a list of applied migrations to find the applied migration with the highest version number. `FindMigrations()` already queries those to check if a migration is already applied. There are several ways to do the check outside of `FindMigrations()` but I did not want to make too many unrelated changes which might be in conflict with other PRs, e.g. changing the return values of `FindMigrations()` would be in conflict with PR #438.

philip-hartmann: I put the strict version check in `Migrate()` and applied some small changes. `FindMigrations()` already returns a sorted list of migrations so we only need to check the first pendig migration with the highest version number from all applied migrations.

pkg/dbmate/db.go
philip-hartmann: Changed the error message to the following:

```
migration `%s` is out of order with already applied migrations, the version number has to be higher than the applied migration `%s` in --strict mode
```

Rutulpatel7077 (approved):

philip-hartmann:
@amacneil my last changes are from 3 months ago without any feedback/review so far, is there anything else open about this PR?

pkg/dbmate/db.go
```diff
@@ -309,17 +312,33 @@ func (db *DB) Migrate() error {
...
+				highestAppliedMigrationVersion = migration.Version
+			}
+		} else {
+			pendingMigrations = append(pendingMigrations, migration)
+		}
+	}
+
+	if len(pendingMigrations) == 0 {
+		return nil
+	}
```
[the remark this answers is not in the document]
dossy: @philip-hartmann Sorry about only bringing this up now, but I thought I had mentioned it months ago when I first looked at the PR but realize I hadn't.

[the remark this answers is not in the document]
philip-hartmann: I cannot think of many scenarios where `db.DumpSchema()` would be necessary if we do not apply any migrations but yes it breaks backwards-compatibility in those scenarios. The `db.DumpSchema()` call can be fixed though.

The `db.openDatabaseForMigration()` call is a little bit more tricky. `db.FindMigrations()` and `db.DumpSchema()` both call `drv.Open()` along the line and return an error if the database cannot be opened. So if the database cannot be opened the `db.FindMigrations()` call would already throw an error earlier in this function.

I am still confident that `--strict` can be introduced without any breaking changes.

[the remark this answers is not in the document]
philip-hartmann: @dossy I've applied the changes to ensure that `db.DumpSchema()` is still called.

pkg/dbmate/db.go
```diff
@@ -324,51 +324,49 @@ func (db *DB) Migrate() error {
...
+				err = execMigration(sqlDB)
+			}
 
-		if err != nil {
-			return err
+			if err != nil {
+				return err
+			}
 		}
 	}
```
dossy, line 371: Instead of including all of this inside the outer `if` statement starting on line 327, would just using this work, instead:

```
	if len(pendingMigrations) > 0 && db.Strict && pendingMigrations[0].Version <= highestAppliedMigrationVersion {
		return fmt.Errorf("migration `%s` is out of order with already applied migrations, the version number has to be higher than the applied migration `%s` in --strict mode", pendingMigrations[0].Version, highestAppliedMigrationVersion)
	}
```

This reduces the amount of code executed inside the conditional, and achieves the same effect, I believe.  What do you think?

dossy (changes requested):

pkg/dbmate/db.go
philip-hartmann: It would work yes, as the condition is only really necessary for that part. But everything afterwards within the if statement benefits too, right? We do not need to open the database if we do not do any migrations. But I am not sure how multiple calls to `drv.Open()` are handled or if they are optimized by go either way. The for loop will most definitely be optimized though and does not need to be part of the if statement.

dossy: It's not a matter of multiple calls to `drv.Open()` but the fact that moving the call to `db.openDatabaseForMigration()` inside the `if` condition means that any side effects of `db.openDatabaseForMigration()` (such as creating, or at least attempting to create, the migrations table) also won't happen.

The original code, when there are no migrations to apply, will still create, or attempt to create, the migrations table.

With your code as it is currently, that won't happen.  This is a subtle but meaningful difference in behavior, that if someone else is depending on, this change will break on them.

Unless there's a good reason to introduce a backwards-incompatible breaking change, I would prefer not to.

philip-hartmann: Right, I totally missed that `db.FindMigrations()` only checks if the migration table exists but does not create the migration tables like `db.openDatabaseForMigration()`. I fixed the if statement. Thanks for pointing that out 👍

dossy (changes requested):
The code looks good to me now, the only thing left is to add a few new test cases to `pkg/dbmate/db_test.go`:

* while strict=false, allows applying migration out of order, the expected behavior before this change was introduced
* while strict=true, allows applying migration in order when no previous migrations have been applied yet
* while strict=true, allows applying migration in order when at least one previous migration has been applied and the new migration is in the correct order
* while strict=true, raises an error when trying to apply a migration out of order

philip-hartmann:
@dossy I've added 2 tests which should cover the 4 cases. Can you please check them?

dossy (approved):
Code reviewed, tests passing: unless someone objects, I think this is good to merge now.

dossy:
@amacneil If there's something still unresolved that you'd like to see changed in this PR, could you add a note here letting us know?  Otherwise, I'd like to clear your change requested bit on this PR and merge it.

Issue #445, opened 2023-05-17
Postgres Schemas with special characters have DBMate create migration tables then cannot find it again


mijamo:
**Description**

- Version: 2.3.0
- Database: Postgres 14
- Operating System: Mac OS

**Steps To Reproduce**
- Create a schema with the name `my-schema`
- Use `search_path=my-schema` or `DBMATE_MIGRATIONS_TABLE="my-schema.schema_migrations"`
- Run `dbmate migrate`
- The migration table is created as expected
- Run `dbmate status`
- No applied migration is detected

**Expected Behavior**
DBMate should find the migration table correctly

I looked quickly at the code (I am not very experienced in Go) and I suspect that the issue comes from https://github.com/amacneil/dbmate/blob/a336dbbd925b1234ccb2a2e43faaed4e76980520/pkg/driver/postgres/postgres.go#L230 where it might be querying Postgres with the already quoted name, so querying for `"my-schema"` instead of `my-schema` and thus des not find the table?

PR #446, opened 2023-05-17
Postgres: Add support for schemas containing special characters

alfredringstad:
As stated in #445, the MigrationsTableExists function does not properly query when the schema contains special characters, since it is then querying for the quoted value.

This PR makes it use unquoted values instead, and adds tests for this behaviour (the tests will fail on the current main branch).

### DIFF
--- a/pkg/driver/postgres/postgres.go
+++ b/pkg/driver/postgres/postgres.go
@@ -221,11 +221,12 @@ func (drv *Driver) DatabaseExists() (bool, error) {
 
 // MigrationsTableExists checks if the schema_migrations table exists
 func (drv *Driver) MigrationsTableExists(db *sql.DB) (bool, error) {
-	schema, migrationsTable, err := drv.quotedMigrationsTableNameParts(db)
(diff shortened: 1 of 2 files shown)


gregwebs:
awesome, thank you for the contribution!
@dossy let's get this merged.

dossy (approved):
Nice, this looks good.

While I think we can merge this PR as-is, I would love to see three more test cases, where `drv.migrationsTableName` is set to a `schema.tablename` where `schema` contains a character that would require quoting in one case, `tablename` contains a character that would require quoting in another case, and where both `schema` and `tablename` contain characters that require quoting in the third case.

This is purely for completeness, and like I said, I think this PR can definitely be merged as-is, but if @alfredringstad agrees and adds these three additional tests, I would be happy to wait on merging until they're added.

alfredringstad:
Thanks! ☺️

mijamo:
Hello! Would it be possible to get a patch release with those changes included @dossy ? I am unsure about the release process so forgive me if that was already planned

amacneil:
Done!

PR #459, opened 2023-07-13
Add further common Postgres Unix socket URL examples

malteneuss:
Add more and shorter URL syntax examples when no password is needed for the Unix socket connection.

feel free to edit as you please.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,19 @@ A `socket` or `host` parameter can be specified to connect through a unix socket
 DATABASE_URL="postgres://username:password@/database_name?socket=/var/run/postgresql"
 ```
 
+For passwordless authentication like Postgres [peer auth-method](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html) the URL can be shortened:
+
+```sh
+DATABASE_URL="postgres://username/database_name?socket=/var/run/postgresql"
+```
+
+Additionaly, when the username and database name are identical the URL can be shortened further:
+
+```sh
+DATABASE_URL="postgres:///database_name?socket=/var/run/postgresql"
+```
+
+
 A `search_path` parameter can be used to specify the [current schema](https://www.postgresql.org/docs/13/ddl-schemas.html#DDL-SCHEMAS-PATH) while applying migrations, as well as for dbmate's `schema_migrations` table.
 If the schema does not exist, it will be created automatically. If multiple comma-separated schemas are passed, the first will be used for the `schema_migrations` table.
 


README.md
```diff
@@ -210,6 +210,19 @@ A `socket` or `host` parameter can be specified to connect through a unix socket
 DATABASE_URL="postgres://username:password@/database_name?socket=/var/run/postgresql"
 ```
 
+For passwordless authentication like Postgres [peer auth-method](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html) the URL can be shortened:
+
+```sh
+DATABASE_URL="postgres://username/database_name?socket=/var/run/postgresql"
```
amacneil, line 216: Does this work? I would have expected needing at least `postgres://username@/database_name?socket=/var/run/postgresql` (with `@`, to differentiate it from a hostname)

README.md
```diff
@@ -210,6 +210,19 @@ A `socket` or `host` parameter can be specified to connect through a unix socket
...
+```
+
+Additionaly, when the username and database name are identical the URL can be shortened further:
```
amacneil, line 219: Huh, this is also surprising to me. I guess it's postgres-specific behavior. It's not something dbmate does.

README.md
malteneuss: Yes, i tested it with Postgres 14.9.

README.md
malteneuss: Same, i tested it with Postgres 14.9.

README.md
dossy: @amacneil This only works with a local connection with the [`peer` authentication method](https://www.postgresql.org/docs/current/auth-peer.html), because in that specific mode, Postgres isn't using the hostname or username from the connection string, as it's connecting locally (so hostname is irrelevant) and `peer` auth relies on Postgres getting the OS username, so the username from the connection string is irrelevant.

In other words, this works:

```
postgres@e8f183dbe0dd:/tmp$ ./dbmate -u postgres:///dbmate_test status
[ ] 20231115033218_test.sql

Applied: 0
Pending: 1
```

Notice that I'm doing this as user `postgres`, which happens to be a valid database user.  If I try this as user `root` that isn't a valid user in this Postgres database, we get this error:

```
root@e8f183dbe0dd:/tmp# ./dbmate -u postgres:///dbmate_test status
Error: pq: role "root" does not exist
```

README.md
dossy: It actually has nothing to do with the username and database name being identical.  When relying on Postgres peer authentication method, specifying the username and hostname in the connection string are unnecessary:

Check this out:

```
postgres@e8f183dbe0dd:/tmp$ ./dbmate -u postgres://hostnameIsIgnoredWhenSocketIsUsed/dbmate_test?socket=/var/run/postgresql status
[ ] 20231115033218_test.sql

Applied: 0
Pending: 1

postgres@e8f183dbe0dd:/tmp$ ./dbmate -u postgres:///dbmate_test?socket=/var/run/postgresql status
[ ] 20231115033218_test.sql

Applied: 0
Pending: 1
```

I think adding these shorthand Postgres database URL examples to the documenation might be handy, but the current PR's explanation for why they work in the README are not accurate.

PR #466, opened 2023-08-25
Update README.md about percent encoding DATABASE_URL

kmchmk:
Passwords with special characters must be URL encoded (percent encoded) when using the DATABASE_URL variable. Otherwise an error might pop up from net/url package when parsing. So it is probably better to mention that in the documentation.


amacneil:
thanks!

PR #469, opened 2023-09-01
Add load command to import schema

s1owjke:
I'm suggesting to add the ability to import a schema from file as is done in [Active Record Migrations](https://guides.rubyonrails.org/active_record_migrations.html)

**Potential Benefits:**
- No extra dependencies and simpler configuration, to setup environments for tests it's enough just to import the schema.sql, instead of adding extra utilities for specific database to do that
- Small performance boost on projects with a large number of migrations, not really relevant for production, but in ci/cd it can save some time


pkg/dbmate/db.go
```diff
@@ -148,25 +153,41 @@ func (db *DB) Wait() error {
 	return db.wait(drv)
 }
 
-// CreateAndMigrate creates the database (if necessary) and runs migrations
+// CreateAndMigrate creates the database (if necessary), load schema.sql file
+// to the database (if autoload enabled) and then apply all pending migrations
 func (db *DB) CreateAndMigrate() error {
```
s1owjke, line 158: If we constantly update the `schema.sql`, then we don't need to apply migrations after the import

pkg/dbmate/db.go
```diff
@@ -215,7 +236,60 @@ func (db *DB) DumpSchema() error {
 	}
 
 	// write schema to file
-	return os.WriteFile(db.SchemaFile, schema, 0o644)
+	if err = os.WriteFile(db.SchemaFile, schema, 0o644); err != nil {
+		return err
+	}
+
+	// in prune mode we delete all migrations from disk after dump
+	if db.Prune {
```
s1owjke, line 244: To implement squashing for migrations after dump we simply clean migrations folder, as that dump already represents latest state

pkg/dbmate/db_test.go
```diff
@@ -38,6 +39,29 @@ func newTestDB(t *testing.T, u *url.URL) *dbmate.DB {
 	return db
 }
 
+func copyFile(from string, to string) error {
```
s1owjke, line 42: Probably not the best place for that utility, but it's used only here

pkg/dbmate/db_test.go
```diff
@@ -194,8 +449,16 @@ func testWaitBefore(t *testing.T, verbose bool) {
 	db.WaitInterval = time.Millisecond
 	db.WaitTimeout = 5 * time.Millisecond
 
+	// create custom schema file directory
```
s1owjke, line 452: Because last dump modify contents of `testdata/schema.sql`

pkg/dbmate/db_test.go
```diff
@@ -341,22 +583,15 @@ func TestRollback(t *testing.T) {
...
 
-			// verify migration
+			// verify results
 			sqlDB, err := drv.Open()
 			require.NoError(t, err)
 			defer dbutil.MustClose(sqlDB)
 
-			var applied []string
-			rows, err := sqlDB.Query("select version from schema_migrations order by version asc")
+			// check applied migrations
```
s1owjke, line 591: For consistency and readability

pkg/dbmate/db_test.go
```diff
@@ -273,45 +547,13 @@ func TestMigrate(t *testing.T) {
...
-
-			err = sqlDB.QueryRow("select count(*) from users").Scan(&count)
-			require.NoError(t, err)
-			require.Equal(t, 1, count)
-		})
-	}
-}
-
-func TestUp(t *testing.T) {
-	for _, u := range testURLs() {
```
s1owjke, line 290: As dumps database specific we need to have dump for each database or perform this test on a known base (I chose this option)

.gitignore
```diff
@@ -4,7 +4,6 @@
 /.cache
 /db
 /dbmate
-/testdata/db/schema.sql
```
dossy, line 7: The schema dumped as a result of test runs should not be commited to the repo.  I think this commit needs to be reverted.

pkg/dbmate/db.go
```diff
@@ -148,25 +153,41 @@ func (db *DB) Wait() error {
...
+		migrationsTableExists, err := drv.MigrationsTableExists(sqlDB)
+		if err == nil && !migrationsTableExists {
+			if err := db.loadSchema(sqlDB); err != nil {
+				return err
+			}
+		}
+	}
+
+	// apply any pending migrations
+	return db.migrate()
```
dossy, line 190: I don't understand why you couldn't still just invoke `db.Migrate()` here, and why you split it into `db.migrate()` and `db.Migrate()` just to move the `len(migrations) == 0` test, at the cost of forcing every call to `db.Migrate()` to do double the work of  `db.FindMigrations()`, once in `db.Migrate()` and a second time in `db.mgirate()`.  This feels like a mistake.

dossy:
As I've never worked on a project where this was ever done, I don't have an opinion on how useful this kind of schema squashing really is.  I'll let others who have experience with this functionality and its merits drive the review.

amacneil:
I think there is quite a lot happening in this PR.

I would be in favor of adding a `dbmate load` command to load the schema.sql file, so you could use that in CI etc if you want. It's pretty much the same as running `psql` so maybe not super useful, but it also seems relatively straightforward.

Regarding "squashing migrations", my opinion is that you can already do that today. Just take the schema.sql contents, put that into a new migration, make sure you're using `IF EXISTS` guards in each of the statements, and delete the old migrations. Dbmate should be perfectly happy with you doing this - you will have some extra entries in your `schema_migrations` table but they won't affect anything.

I don't understand the other changes `prune` and `--autoload-schema`. I prefer we don't make any breaking changes to `dbmate up` or any new features that add complexity for people using the tool.

.gitignore
s1owjke: Since I originally included the schema file in the repository code this was required to add into repository

pkg/dbmate/db.go
s1owjke: You're right, there was a mistake

s1owjke:
@amacneil I removed the squashing functionality completely, you are right that this can be achieved in another way, and if squashing happens in a project, it doesn't happen too often

Regarding the load schema - In ci/cd this should make flow a bit easier, e.g. mysql cli does not accept uri and expects a different format - separate host, port, login and password

Without additional flags in the current api it is not possible to achieve backwards compatibility, but I absolutely agree that there is no point in complicating the current one, let's keep it simple

pkg/dbmate/db.go
```diff
@@ -218,6 +218,41 @@ func (db *DB) DumpSchema() error {
...
+
+	sqlDB, err := drv.Open()
+	if err != nil {
+		return err
+	}
+	defer dbutil.MustClose(sqlDB)
+
+	_, err = os.Stat(db.SchemaFile)
+	if err != nil {
+		return nil
```
amacneil, line 236: shouldn't this return a wrapped error?

amacneil:
Sounds good! 

@dossy want to do a code review here?

pkg/dbmate/db.go
dossy: Agree: if we fail to `os.Stat()` the file, we should raise an error and not terminate successfully.

pkg/dbmate/db_test.go
```diff
@@ -175,6 +175,60 @@ func TestAutoDumpSchema(t *testing.T) {
...
+	appliedMigrations, err := drv.SelectMigrations(sqlDB, -1)
+	require.NoError(t, err)
+	require.Equal(t, map[string]bool{"20200227231541": true, "20151129054053": true}, appliedMigrations)
+
+	// users and posts tables have been created
+	var count int
+	err = sqlDB.QueryRow("select count(*) from users").Scan(&count)
+	require.Nil(t, err)
+	err = sqlDB.QueryRow("select count(*) from posts").Scan(&count)
+	require.Nil(t, err)
```
dossy, line 229: I believe that these may be functionally equivalent, but would prefer you use `require.NoError(t, err)` here instead of `require.Nil(t, err)`, as it expresses the intent and expectation more clearly.

dossy (changes requested):

pkg/dbmate/db_test.go
s1owjke: I fixed it to make it cleaner, they are the same internally, note that both forms are now used in tests, so I just wrote in the same style.

pkg/dbmate/db.go
```diff
@@ -233,7 +234,7 @@ func (db *DB) LoadSchema() error {
 
 	_, err = os.Stat(db.SchemaFile)
 	if err != nil {
-		return nil
+		return ErrSchemaNotFound
 	}
```
dossy, line 238: Personally, I would prefer you just return `err` here, because you are assuming that `os.IsNotExist(err)` is true here with the error message you're returning, but that may not necessarily be the case.

Unfortunately, it's hard to automate a test around other situations where `os.Stat()` may return an error because if you're [using a Mac to run Docker, then permission handling on volume mounts makes this difficult](https://github.com/docker/for-mac/issues/2657).  We can use a container-native filesystem mountpoint like `/tmp`, though:

```sh
$ docker compose run --rm dev /bin/sh

$ mkdir /tmp/inaccessible
$ touch /tmp/inaccessible/unreadable
$ chown -R 0:0 /tmp/inaccessible
$ chmod 0 /tmp/inaccessible

$ su nobody -s /bin/sh -c "dist/dbmate --url $POSTGRES_TEST_URL --schema-file /tmp/inaccessible/unreadable load"

## Output:

Error: stat /tmp/inaccessible/unreadable: permission denied
```

That's what `dbmate` will output if you just return `err` up to the caller.  The built-in error messages are concise and specific and clear:

```sh
$ su nobody -s /bin/sh -c "dist/dbmate --url $POSTGRES_TEST_URL --schema-file /tmp/nonexistent load"
Error: stat /tmp/nonexistent: no such file or directory
```

dossy (changes requested):

pkg/dbmate/db.go
s1owjke: I agree with you, it will be clearer for the user in this case than just general error.

And special thanks for explaining how to simulate it in docker on Mac.

amacneil:
@dossy want to re-review? then we can merge

dossy (approved):
LGTM now, thanks!

PR #470, opened 2023-09-04
libsql: Add initial support

Ty3uK:
libsql code heavely based on sqlite code.

Database drop was written manually because you cannot remove remote database.


amacneil:
Never heard of libQSL before. The [github page](https://github.com/libsql/libsql) says that they will maintain 100% compatibility with the file format and API, so why is a new driver needed?

Ty3uK:
> Never heard of libQSL before. The [github page](https://github.com/libsql/libsql) says that they will maintain 100% compatibility with the file format and API, so why is a new driver needed?

Because libSQL also allows you to use remote db over http/ws/grpc.

Check [Turso](https://turso.tech/)

pkg/driver/libsql/libsql.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/lib/pq"
+	_ "github.com/libsql/libsql-client-go/libsql" // database/sql driver
+)
+
+func init() {
+	dbmate.RegisterDriver(NewDriver, "libsql")
+	dbmate.RegisterDriver(NewDriver, "http")
+	dbmate.RegisterDriver(NewDriver, "https")
+	dbmate.RegisterDriver(NewDriver, "ws")
+	dbmate.RegisterDriver(NewDriver, "wss")
```
amacneil, line 26: huh? why is it accessible over http + websocket?

Ty3uK: Check this out 🙂

https://github.com/libsql/sqld

emilpriver:
Hey, Just wondering if there is any update on this topic? :D

crvouga:
Is there any updates on this? Will it be added soon?

pkg/driver/libsql/libsql.go
dossy: Registering the very generic `http`, `https`, `ws` and `wss` to this driver is probably going to bite us in the future.  I would rather see us register `libsql+http`, `libsql+https`, `libsql+ws` and `libsql+wss` and peel off the `libsql+` prefix inside `ConnectionString()`.

pkg/driver/libsql/libsql.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	migrations, err := drv.schemaMigrationsDump(db)
+	if err != nil {
+		return nil, err
+	}
+
+	schema = append(schema, migrations...)
+	return dbutil.TrimLeadingSQLComments(schema)
+}
```
dossy, line 141: This doesn't currently work:

```
# dist/dbmate -u $LIBSQL_TEST_URL -d testdata/db/migrations dump
Error: Error: 'tls' usage forbidden. Please use 'WithTls' option instead
```

PR #471, opened 2023-09-06
Add wait-for-current command, which waits until the database is migrated by a third party.

sgielen:
This way, we can delay the start of an application until its expected schema is present.

---

In our setup, we use `dbmate migrate` in a Kubernetes Job to perform the schema migration whenever we deploy a new version, at the same time as updating the (multiple) Kubernetes Deployments to run the new version of the application.

We take care to keep our migrations backwards compatible for at least one release, so schema version 16 can be used by backend version 15. The schema is of course not forwards compatible: schema version 15 cannot be used by backend version 16. Because it is difficult to express a dependency between the Deployments and the migration Job, it sometimes happens that backend version 16 starts up while the schema migration for version 16 is still ongoing, meaning backend 16 runs on schema 15 and starts giving RPC errors.

In order to resolve this, I would like to have backend 16 wait until its migration job is succeeded, by polling the database until the migration is completed.

This PR adds `dbmate wait-for-current`, which waits until the database schema has at least all migrations applied to it. It does not apply any migrations itself. This way, backend 16 only truly starts when migration 16 has completed, resolving the issue.

### DIFF
--- a/main.go
+++ b/main.go
@@ -206,6 +206,21 @@ func NewApp() *cli.App {
 				return db.Wait()
 			}),
 		},
+		{
+			Name:  "wait-for-current",
+			Usage: "Wait for the database schema to become current",
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:    "quiet",
+					Aliases: []string{"v"},
+					Usage:   "don't output any text",
+				},
+			},
+			Action: action(func(db *dbmate.DB, c *cli.Context) error {
+				quiet := c.Bool("quiet")
+				return db.WaitForCurrent(quiet)
+			}),
+		},
 	}
 
 	return app
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -553,3 +553,31 @@ func (db *DB) Status(quiet bool) (int, error) {
 
 	return totalPending, nil
 }
+
+// WaitForCurrent waits until all migrations are applied before exiting.
+func (db *DB) WaitForCurrent(quiet bool) error {
+	for {
+		results, err := db.FindMigrations()
+		if err != nil {
+			return err
+		}
+
+		var missing []string
+		for _, res := range results {
+			if !res.Applied {
+				missing = append(missing, res.FileName)
+			}
+		}
+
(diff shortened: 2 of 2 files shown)


sgielen:
@amacneil bump - any thoughts about this addition? :-)

sgielen:
@amacneil any chance you might have time to look at this PR sometime soon?

main.go
```diff
@@ -206,6 +206,21 @@ func NewApp() *cli.App {
 				return db.Wait()
 			}),
 		},
+		{
+			Name:  "wait-for-current",
+			Usage: "Wait for the database schema to become current",
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:    "quiet",
+					Aliases: []string{"v"},
```
dossy, line 215: I would expect the flag `-v` to be verbose, and `-q` to be quiet.  Do you agree?

dossy (commented):
I understand why this functionality can be useful, but want to point out that this can be implemented today with a one-liner shell wrapper:

```
while test $(dist/dbmate -u $MYSQL_TEST_URL -d testdata/db/migrations status --exit-code --quiet; echo $?) -gt 0; do echo "Waiting..."; sleep 5; done
```

Rather than introducing a new `wait-for-current` subcommand, maybe adding a `--wait-until-current <seconds>` subcommand option could be added to `status` instead, where `<seconds>` is an optional number of seconds to wait before exiting, so that this can be used in a way that doesn't necessarily wait forever, while the default is `-1` (forever), and the current behavior of status is effectively `--wait-until-current 0` (wait 0 seconds, or don't wait).

What do you think about this, @sgielen?

amacneil:
@sgielen sorry, I have had not had much time to dedicate to this project lately

@dossy thanks for the review

I agree with @dossy's comment, I think a bash one-liner here is probably better than a new command. I don't think it is obvious from the name `dbmate wait-for-current` what it would do. You can even make the bash command simpler:

```sh
while ! dbmate status --exit-code; do sleep 1; done
```

Given that, I'll close the PR, but if I'm missing something here please feel free to correct me.

Issue #444, opened 2023-05-14
Dbmate Unable to Create SchemaMigrations Table


dilutedev:
**While running `dbmate up` I get this error Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '"schema_migrations"' at line 1**

- Version: 8
- Database: mysql
- Operating System: macos

**Steps To Reproduce**
`Just run dbmate up with mysql:v8

**Should migrate data but it logs that error**

ebuckthal:
My guess is that you have [ANSI_QUOTES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes) SQL mode enabled. It's not a default flag in MYSQL 8 itself, but I've noticed some database providers turn it on by default.

I believe `dbmate` should use back ticks instead of double quotes in these instances

prayog-exc:
Did you ever find a solution to this @dilutedev ? I'm running into the same issue.

dilutedev:
> Did you ever find a solution to this @dilutedev ? I'm running into the same issue.

No, I didnt

dilutedev:
> Did you ever find a solution to this @dilutedev ? I'm running into the same issue.

No, I didnt

toru:
I came across this problem with mysqld (8.0.30) that was configured with the `ANSI_QUOTES` sql_mode. This issue is not limited to just the `schema_migrations` table, and it affects most commands because `schema_migrations` related queries are integral to dbmate.

Here is the explanation of the problem, and how I solved it. First, a demonstration to show that the ANSI_QUOTES sql_mode was enabled.

```
> SELECT @@sql_mode;
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| @@sql_mode                                                                                                                                                                      |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE,ONLY_FULL_GROUP_BY,ANSI,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.001 sec)

[...]

PR #475, opened 2023-09-14
MySQL: Add ANSI_QUOTES support to MigrationsTableExists()

toru:
This PR addresses #444

## Problem

Dbmate currently fails to function against MySQL instances enabled with the [ANSI_QUOTES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes) sql_mode.

My initial investigation report [can be found here](https://github.com/amacneil/dbmate/issues/444#issuecomment-1719787799).

## Why

The ANSI_QUOTES mode does not allow double-quote characters to be used for expressing string literals. Therefore ANSI_QUOTES-enabled MySQL rejects the following query produced by Dbmate's MySQL driver:

```
SHOW TABLES LIKE "schema_migrations"
```
From: https://github.com/amacneil/dbmate/blob/6245b8c/pkg/driver/mysql/mysql.go#L236

## Fix & Thought Process

MySQL driver's [quotedMigrationsTableName()](https://github.com/amacneil/dbmate/blob/6245b8c/pkg/driver/mysql/mysql.go#L316) complies with ANSI_QUOTES. Therefore I initially attempted to replicate other command implementations over to [MigrationsTableExists()](https://github.com/amacneil/dbmate/blob/6245b8c/pkg/driver/mysql/mysql.go#L234) like so:

```diff
-       err := db.QueryRow(fmt.Sprintf("SHOW TABLES LIKE \"%s\"",
-               drv.migrationsTableName)).
-               Scan(&match)
+       err := db.QueryRow("show tables like ?", drv.quotedMigrationsTableName()).Scan(&match)
```

This would be ideal for two reasons:

* Code-base consistency
* Delegating string interpolation to the sql package

However the query produced by the above code was considered invalid by MySQL (with and without ANSI_QUOTES). It appears that the LIKE predicate is picky about how the condition is quoted. Fair enough, I went a step further and tried this code instead:

```diff
- err := db.QueryRow("show tables like ?", drv.quotedMigrationsTableName()).Scan(&match)
+ err := db.QueryRow("show tables like '?'", drv.quotedMigrationsTableName()).Scan(&match)
```

Unfortunately, this time, the sql package's string interpolation was negatively affected. It seems to not appreciate the quotes around the `?` interpolation character.

```
Error: sql: expected 0 arguments, got 1
```

At this point I understood the agony of the person who last worked on this part of the code, and why MySQL driver's `MigrationsTableExists()` has an inconsistent `fmt.Sprintf` based implementation. One code consistency I was able to pick up was lower-casing the entire query.

The final code change may appear lazy but in fact a lot of thought went into it. Cheers!

### DIFF
--- a/pkg/driver/mysql/mysql.go
+++ b/pkg/driver/mysql/mysql.go
@@ -233,7 +233,7 @@ func (drv *Driver) DatabaseExists() (bool, error) {
 // MigrationsTableExists checks if the schema_migrations table exists
 func (drv *Driver) MigrationsTableExists(db *sql.DB) (bool, error) {
 	match := ""
-	err := db.QueryRow(fmt.Sprintf("SHOW TABLES LIKE \"%s\"",
+	err := db.QueryRow(fmt.Sprintf("show tables like '%s'",
 		drv.migrationsTableName)).
 		Scan(&match)
 	if err == sql.ErrNoRows {


toru:
Apologies. My example code diff was incorrect. I am cursed with a very strong muscle memory of typing: `show create table`.

dossy (approved):
This is still not ideal as we should really escape values, but considering that even [database/sql hasn't provided a driver-level escaping mechanism](https://github.com/golang/go/issues/18478) in nearly 7 years suggests that we will either have to implement one ourselves, or just 🙈 and hope for the best.

toru:
Appreciate your understanding @dossy! Agreed, the current structure is not ideal. These edge-case code paths can quickly become an attack vector. I suppose we can make peace today by calling it necessary evil. It would indeed be cool to revise the code at some point, like you say, even if it means bringing the escaping responsibility to Dbmate's layer.

toru:
For posterity, what prompted me to submit this PR was DigitalOcean's managed MySQL service. At the time of this comment, their instances are bootstrapped with the `ANSI_QUOTES` SQL option.

PR #488, opened 2023-10-17
Add Redshift support

aterekhov-plr:
This small fix checks that the database is Redshift and make client-side quoting instead of server quoting.


dossy:
I don't have a Redshift cluster handy to test this PR against, so I'm labeling this PR as `help wanted` in case someone else who has access to a Redshift cluster would like to test this.  Looking at the PR, it's unclear why this change is even necessary, as the server-side quoting should work for ordinary Postgres as well as Redshift.

@aterekhov-plr what is the error you get when trying to use dbmate with Redshift, that this PR fixes?  In other words: how can someone else verify that there's a problem that is fixed by this PR?

gregwebs:
@dossy the error was shown here: https://github.com/amacneil/dbmate/discussions/361

aterekhov-plr:
@dossy If you have time to test this PR and help with the release, I could try to provide access to a temporary Redshift cluster for you.

dossy:
@aterekhov-plr That would be helpful, if you could provide temporary access.  I can find the time next week to work on this.  My email is listed on my GitHub profile, and so are [my PGP/GPG keys](https://github.com/dossy.gpg).

amacneil:
Looking at how other software handles redshift (e.g. [golang-migrate](https://pkg.go.dev/github.com/golang-migrate/migrate/database/redshift), [goose](https://github.com/pressly/goose/blob/master/internal/dialect/dialectquery/redshift.go), official [redshift go package](https://pkg.go.dev/github.com/aws/aws-sdk-go/service/redshift)), I think it might make more sense to add an entirely new `redshift://` driver (even if it mostly inherits from the postgres driver). That way we can allow the user to be explicit rather than implicit auto-detection.

pkg/dbmate/version.go
```diff
@@ -1,4 +1,4 @@
 package dbmate
 
 // Version of dbmate
-const Version = "2.6.0"
```
amacneil, line 4: can you remove this please, we update the version only when making releases

aterekhov-plr:
> Looking at how other software handles redshift (e.g. [golang-migrate](https://pkg.go.dev/github.com/golang-migrate/migrate/database/redshift), [goose](https://github.com/pressly/goose/blob/master/internal/dialect/dialectquery/redshift.go), official [redshift go package](https://pkg.go.dev/github.com/aws/aws-sdk-go/service/redshift)), I think it might make more sense to add an entirely new `redshift://` driver (even if it mostly inherits from the postgres driver). That way we can allow the user to be explicit rather than implicit auto-detection.

I agree that creating a new driver is the best approach to implement Redshift support. I began working on it but later realized that the difference between the Redshift driver and the PostgreSQL driver would only be a few lines of code. I might revisit it later, but I currently need some assistance with tests. There have been issues with ports (5432/5439) if use pg, and it's unclear what to use in the docker-compose setup.

dossy:
@aterekhov-plr, @amacneil -- I started down the path of trying to create an entire separate `redshift` driver package, but as golang doesn't have OO where I could "subclass" the `postgres` driver, it resulted in mostly a duplicated copy of the code.  I opted to register `redshift` as a driver prefix, instead.

I created a variant of one of the PG tests that I could run against the Redshift instance that @aterekhov-plr provided (thanks!) so that I could demonstrate that the original driver implementation "failed" and introducing Andrei's fix makes the test pass, so I'm happy with this.

In order to run the Redshift test, set the `REDSHIFT_TEST_URL` environment variable with a `redshift://` connection string to a Redshift cluster, before running the tests.

Let me know what you both think, but I think this is good to go, unless either of you feel otherwise?

pavel-popov-playrix:
@dossy @amacneil do you need a temporary redshift cluster for testing or can we terminate it? Cos right now we're spending extra money for it, so if you no longer need the cluster we will terminate it.

dossy:
@pavel-popov-playrix I was able to complete my testing using it, so I'm okay with terminating it.  Thank you so much for your generosity.

.gitignore
```diff
@@ -9,3 +9,4 @@
 /vendor
 dist
 node_modules
+docker-compose.override.yml
```
amacneil, line 12: nit: I try to keep this file alphabetically sorted

pkg/driver/postgres/postgres.go
```diff
@@ -115,10 +124,12 @@ func (drv *Driver) openPostgresDB() (*sql.DB, error) {
...
 
-	// connect to postgres database
-	postgresURL.Path = "postgres"
+	// connect to postgres database, unless this is a Redshift connection
+	if drv.databaseURL.Scheme != "redshift" {
+		postgresURL.Path = "postgres"
+	}
 
-	return sql.Open("postgres", postgresURL.String())
+	return sql.Open("postgres", connectionString(postgresURL))
```
amacneil, line 132: why is this necessary? we already ran it through `connectionString()` above

pkg/driver/postgres/postgres.go
```diff
@@ -115,10 +124,12 @@ func (drv *Driver) openPostgresDB() (*sql.DB, error) {
 		return nil, err
 	}
 
-	// connect to postgres database
-	postgresURL.Path = "postgres"
+	// connect to postgres database, unless this is a Redshift connection
+	if drv.databaseURL.Scheme != "redshift" {
+		postgresURL.Path = "postgres"
```
amacneil, line 129: If we don't change the root DB here, does that mean that `create` / `drop` / `dump` etc don't work at all?

In that case maybe it is better to give a specific error message rather than silently failing?

.gitignore
amacneil, line 12: nit: I try to keep this file alphabetically sorted

amacneil (commented):
Looking good, I added a few comments/questions

pkg/driver/postgres/postgres.go
dossy: On standard out-of-the-box Redshift, there is no database named `postgres`.

The [documentation](https://docs.aws.amazon.com/redshift/latest/dg/r_DROP_DATABASE.html) specifies that you cannot drop the `dev`, `padb_harvest`, `template0`, `template1`, or `sys:internal` databases, nor the current database.  So, for Redshift, we could try connecting to the `sys:internal` database in order to drop the target database, as long as it's not one of the listed databases.

I don't think we'll silently fail.  We should get an error from Redshift that the database cannot be dropped, most likely because it's the current database.

pkg/driver/postgres/postgres.go
dossy: Good catch, not sure why I thought this was necessary.

pkg/driver/postgres/postgres.go
dossy: @aterekhov-plr How important is it for dbmate to be able to drop/create the Redshift database?

dossy: @amacneil How do you feel about this?  I like @aterekhov-plr's approach of moving forward as-is, and if anyone needs the functionality, they can open an issue and we can sort it out then?

amacneil: I don't think we need drop/create functionality for redshift, as long as the error message is clear.

amacneil (approved):
Thank you!

Issue #274, opened 2022-02-22
Support env.local, et al. or support `--envfile` CLI option


hparra:
dbmate does not support additional `.env` files, e.g. `.env.local` or `.env.development.local` due to https://github.com/joho/godotenv/issues/10

Possible solutions of varying completeness:
- Have [`loadDotEnv`](https://github.com/amacneil/dbmate/blob/519f186c8a3daec086608b2c21677c672eda56db/main.go#L215) also load `.env.local`
- Add `--envfile` option which specifies file for `godotenv.Load()`
- Use a fork of godotenv which implements https://github.com/joho/godotenv/issues/10

msdrigg:
I also would like this issue supported

amacneil:
Happy to accept a PR for this, I think it would only be a few lines of code

msdrigg:
I haven't written any go code, but if this is still sitting open in a week or two, I may give it a try.

Also, if anyone does write a pr, I would definitely prefer the solution of
> Add `--envfile` option which specifies file for `godotenv.Load()`

jahilliard:
I'd love this too.

jahilliard:
Looks like something like `godotenv.Load(".env.local")` is supported.

PR #491, opened 2023-10-22
support --env-file CLI option (#274)

hamza512b:
I filing pr after I looked at this discussion https://github.com/amacneil/dbmate/discussions/354.

This pr add support for ability to use different env files than the default one `.env`


dossy (changes requested):
In addition to my other review comments, there needs to be an addition to [the README.md](https://github.com/hamza512b/dbmate/blob/f5290e5ca52da417904e9af6dedda586c9e66c1e/README.md?plain=1#L140-L141) documenting this new Command Line Option.

hamza512b:
I have applied the suggestions, however I came up with an issue, if we pass multiple .env files to the function `loadDotEnv` (there is no way to do it right now) and some file does not exists how should we handle that?

For now, what I did is return if any file does not exist.

main.go
```diff
@@ -211,15 +224,19 @@ func NewApp() *cli.App {
...
+// load environment variables
+func loadDotEnv(filenames ...string) error {
+	for _, filename := range filenames {
+		if _, err := os.Stat(filename); err != nil {
+			return nil
+		}
 	}
 
-	if err := godotenv.Load(); err != nil {
-		log.Fatalf("Error loading .env file: %s", err.Error())
```
hamza512b, line 221: I removed the dot here becuase the file not always called `.env` now

dossy:
> I have applied the suggestions, however I came up with an issue, if we pass multiple .env files to the function `loadDotEnv` (there is no way to do it right now) and some file does not exists how should we handle that?
> 
> For now, what I did is return if any file does not exist.

We can rely on `godotenv`'s underlying behavior (see [readFile](https://github.com/joho/godotenv/blob/e3b6eee84d15b0fa274078565ca46591f5e08876/godotenv.go#L207-L210)), where it will return an error if a file cannot be opened, because it either doesn't exist or the process lacks sufficient permission to read the file.

main.go
```diff
@@ -211,15 +224,19 @@ func NewApp() *cli.App {
...
+	if err := godotenv.Load(filenames...); err != nil {
+		return fmt.Errorf("Error loading env file: %s", err.Error())
 	}
+
+	return nil
 }
```
dossy, line 240: We can check for the error condition where it is `os.ErrNotExist` and ignore it here, otherwise return the error which our caller can handle as appropriate:

```
// load environment variables from files
func loadDotEnv(filenames ...string) error {
	err := godotenv.Load(filenames...)
	if err != nil {
		var perr *os.PathError
		if errors.As(err, &perr) {
			if errors.Is(perr, os.ErrNotExist) {
				// this is fine, ignore this error
				return nil
			}
		}
		return fmt.Errorf("loading env file(s) %v: %v", filenames, err)
	}

	return nil
}
```

dossy (changes requested):
if you want to support being able to load multiple environment files using repeated `--env-file` arguments, take a look at the `--migrations-dir` flag:

* [How to specify the default value](https://github.com/amacneil/dbmate/blob/9e653e716b92deb9c49deddbcd9c1ea810a29085/main.go#L52-L58)
* [How to retrieve the array of values](https://github.com/amacneil/dbmate/blob/9e653e716b92deb9c49deddbcd9c1ea810a29085/main.go#L247)

Since `loadDotEnv()` already takes an array of strings as its parameter, supporting multiple `--env-file`s is quite easy.

Also, since we're now not only loading `.env`, maybe it makes sense to rename `loadDotEnv` to `loadEnvFiles`?  What do you think?

It would be nice to have a few tests covering this functionality.  Is that something you'd like to try and add?  If not, I can.

hamza512b:
While implementing support for `--env-file` to accept multiple files, I have found an issue with approach I went with. We can not use `app.Before` to load environment variables, because `urfave/cli` will no longer be able use environment variables, because it apparently tries to run `before` function after it had  evaluated the command-line arguments.

One way to deal with this is to manually parces the command-line arguments.

dossy:
> While implementing support for `--env-file` to accept multiple files, I have found an issue with approach I went with. We can not use `app.Before` to load environment variables, because `urfave/cli` will no longer be able use environment variables, because it apparently tries to run `before` function after it had evaluated the command-line arguments.
> 
> One way to deal with this is to manually parses the command-line arguments.

I suspect this is exactly why [`loadDotEnv()` was called from `main`](https://github.com/amacneil/dbmate/blob/9e653e716b92deb9c49deddbcd9c1ea810a29085/main.go#L20) originally.

That's probably the simplest solution, here.

dossy:
Okay, so `urfave/cli` uses the Go `flag` package for its argument parsing, and it [returns a "flag provided but not defined" error](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/flag/flag.go;l=1109-1116;drc=00401835c1d5122a84f0500e48655cd787402f3f) and that's not configurable, so it interrupts flag parsing if an unrecognized flag is present, which we don't want.

In other words, `--env-file .env.local --foo --env-file .env.development.local` will only parse the `--env-file .env.local` and return an error when it reaches `--foo` (as long as it's not defined as a flag), and stops, so `--env-file .env.development.local` will not be parsed.

I think you were right, just implementing a very simple arg parser that just looks for `--env-file <value> [--env-file <value>]` is the way to go.  Let me put that together and add it to this PR.  See [e2618f8](https://github.com/amacneil/dbmate/pull/491/commits/e2618f8cf9514ae19d5f8e7061d8b177a6872d1c).

We should leave the `--env-file` flag definition in the `cli.App` so that it shows up in the help/usage text and parsed when `urfave/cli` parses the command line, but we won't actually do anything with the `--env-file` values it finds.

dossy:
@hamza512b I also added tests to document the behavior of `loadEnvFiles`, in [29e3bcb](https://github.com/amacneil/dbmate/pull/491/commits/29e3bcb0e2061f8fc2f52076bc06e706bd10f5c6).  Please review.  Thanks!

hamza512b:
I have looked through your commits. As far as I can see, I do not see problems with the code.

dossy:
@hamza512b I've updated the discussion https://github.com/amacneil/dbmate/discussions/354, looking for any final feedback from users.  If there's no issues raised, I think this is good to merge.  Thanks!

dossy (approved):

README.md
```diff
@@ -138,6 +138,7 @@ The following options are available with all commands. You must use command line
 
 - `--url, -u "protocol://host:port/dbname"` - specify the database url directly. _(env: `DATABASE_URL`)_
 - `--env, -e "DATABASE_URL"` - specify an environment variable to read the database connection URL from.
+- `--env-file ".env"` - specify an alternate environment variables file(s) to load.
```
amacneil, line 141: might be helpful to give an example of how to specify multiple files, and mention which files are loaded automatically

amacneil:
lgtm. I added a minor comment about the documentation but feel free to merge when ready

README.md
dossy: Agreed, there should be an "Environment Variables" section in the README that explains how dbmate uses them, how `.env` is loaded automatically, and variable precedence: dot-env files, calling process environment, command line arguments.

I created issue #500 for it.  I'll write it up if no one else volunteers, but it's a good opportunity for someone else to contribute, so it's probably okay to leave it up for a while.

Mizaro:
> > I have applied the suggestions, however I came up with an issue, if we pass multiple .env files to the function `loadDotEnv` (there is no way to do it right now) and some file does not exists how should we handle that?
> > For now, what I did is return if any file does not exist.
> 
> We can rely on `godotenv`'s underlying behavior (see [readFile](https://github.com/joho/godotenv/blob/e3b6eee84d15b0fa274078565ca46591f5e08876/godotenv.go#L207-L210)), where it will return an error if a file cannot be opened, such as when it does not exist or when the process lacks sufficient permission to read the file.
> 
> Go's [`os.Open`](https://pkg.go.dev/os#Open) will return an error if the file does not exist, however, and as we want to not halt execution in this case, we need to check the error condition for `os.ErrNotExist` and ignore it if that's the case, otherwise process the error as normal. This way, if the `.env` file, or alternative file specified with `--env-file`, does not exist, we currently won't raise an error, and simply continue, which is equivalent to the file existing but being empty, anyway.

In my humble opinion, if a file does not exist, any tool should tell the user that the input is invalid.
Otherwise, a scenario where the user gives a path to a file without permission can look like everything is fine while the critical environment variable didn't get read.

dossy:
> In my humble opinion, if a file does not exist, any tool should tell the user that the input is invalid.
> Otherwise, a scenario where the user gives a path to a file without permission can look like everything is fine while the critical environment variable didn't get read.

Hi, @Mizaro.  Ordinarily, under normal circumstances, I absolutely agree with you.

However, environment files are a special case and I think ignoring files that do not exist (`os.ErrNotExist`), which is different than when the file exists but the process lacks the permission to read, is useful.

It's a common pattern to specify environment files as a list, such as `[]string{".env.local", ".env"}`, where if the file `.env.local` exists, it is read first, and then `.env` is read, but only new keys in `.env` are incorporated into the environment, not overwriting anything set by `.env.local`.  The desired behavior is, if `.env.local` exists, use it, but if it doesn't, it shouldn't raise an error.

This is currently the behavior implemented in dbmate, which I think is the right behavior.  If you feel it's incorrect, please create an issue or discussion and we can determine if it needs to change.

PR #495, opened 2023-11-15
Add line/column/position details to Postgres migration/rollback query error messages

dossy:
Postgres's database driver returns the character position where a syntax error occurs, so dbmate can use that information to calculate the line and column within the query that the position refers to.                         

This PR supersedes the work from PR #204, simplifying the implementation a bit by minimizing the amount of code change required, plus adding additional tests, including SQL queries with UTF-8 characters present.


amacneil:
Nice work, thank you @dossy and @silbinarywolf

amacneil (approved):

PR #496, opened 2023-11-16
fix/correctly parse crlf line endings

### LINKED ISSUE #213: The first migration is always rolled back on Windows
[20210329163410_initial_migration.txt](https://github.com/amacneil/dbmate/files/6343294/20210329163410_initial_migration.txt)

Synthetic first migration example for demonstration.
```sql
-- migrate:up
CREATE SCHEMA IF NOT EXISTS up;

-- migrate:down
CREATE SCHEMA IF NOT EXISTS down;
```
On Windows, the first migration will always be rolled back and both schemes will appear in the database.
On Linux all is fine.
This behavior was seen only in the first migration. If you add a migration before it, even one consisting of only comments, the `down` schema will not be created.

**! UPDATE:**
I figured out what was causing the poblem. These are Windows line breaks (\r\n). I attach a file created on Windows, with it the same behavior occurs under Linux.

dossy:
- chore: add failing tests that demonstrate the issue
- fix: correct migration parsing regexps

Fixes #213.

### DIFF
--- a/pkg/dbmate/db_test.go
+++ b/pkg/dbmate/db_test.go
@@ -735,3 +735,44 @@ func TestMigrateQueryErrorMessage(t *testing.T) {
 		require.Contains(t, err.Error(), "line: 3, column: 3, position: 29:")
 	})
 }
+
+func TestMigrationContents(t *testing.T) {
+	for _, u := range testURLs() {
+		t.Run(u.Scheme, func(t *testing.T) {
+			t.Run("ensure Windows CR/LF line endings in migration files work", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


amacneil (approved):
🙏 thanks!

feel free to merge PR after you fix the merge conflict

dossy:
@amacneil Since this PR fixes a real bug that's impacting users, what are your thoughts on a timeframe for releasing v2.8.0?

amacneil:
😄 https://github.com/amacneil/dbmate/releases/tag/v2.8.0

PR #499, opened 2023-11-19
Make appending to FLAGS possible when invoking make.

dossy:
Sometimes, it's nice to be able to run just a subset of the test suite.  This change to the `Makefile` enables one to do this:

```sh
$ make test FLAGS='-run "^TestCasePrefix"'
```

Limited usefulness, but without the change to the `Makefile`, this isn't possible, so here's a PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ OUTPUT ?= dbmate
 GOOS := $(shell go env GOOS)
 ifeq ($(GOOS),linux)
 	# statically link binaries to support alpine linux
-	override FLAGS := -tags netgo,osusergo,sqlite_omit_load_extension,sqlite_json  -ldflags '-s -extldflags "-static"' $(FLAGS)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,17 +1,17 @@
...
-
 # default output file
 OUTPUT ?= dbmate
 
 # platform-specific settings
 GOOS := $(shell go env GOOS)
 ifeq ($(GOOS),linux)
 	# statically link binaries to support alpine linux
-	FLAGS := -tags netgo,osusergo,sqlite_omit_load_extension,sqlite_json  -ldflags '-s -extldflags "-static"'
+	override FLAGS := -tags netgo,osusergo,sqlite_omit_load_extension,sqlite_json  -ldflags '-s -extldflags "-static"' $(FLAGS)
```
amacneil, line 11: looks like there was an extra space here in the original file for some reason, may as well fix it now

```suggestion
	override FLAGS := -tags netgo,osusergo,sqlite_omit_load_extension,sqlite_json -ldflags '-s -extldflags "-static"' $(FLAGS)
```

amacneil (approved):

PR #523, opened 2024-03-03
Adds support for BigQuery

mii9000:
re: Support for GCP's Big Query? #522


docker-compose.yml
```diff
@@ -63,3 +65,10 @@ services:
       - zookeeper
     volumes:
       - ./pkg/driver/clickhouse/testdata/cluster_config/ch-cluster-02:/etc/clickhouse-server
+
+  bigquery:
+    image: ghcr.io/goccy/bigquery-emulator:0.4.4
+    ports:
+      - "0.0.0.0:9050:9050"
```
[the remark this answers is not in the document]
mii9000: removed

docker-compose.yml
```diff
@@ -12,13 +12,15 @@ services:
...
       - clickhouse-cluster-02
+      - bigquery
     environment:
       CLICKHOUSE_TEST_URL: clickhouse://clickhouse:9000/dbmate_test
       CLICKHOUSE_CLUSTER_01_TEST_URL: clickhouse://ch-cluster-01:9000/dbmate_test
       CLICKHOUSE_CLUSTER_02_TEST_URL: clickhouse://ch-cluster-02:9000/dbmate_test
       MYSQL_TEST_URL: mysql://root:root@mysql/dbmate_test
       POSTGRES_TEST_URL: postgres://postgres:postgres@postgres/dbmate_test?sslmode=disable
       SQLITE_TEST_URL: sqlite3:/tmp/dbmate_test.sqlite3
+      BIGQUERY_TEST_URL: bigquery://test/dbmate_test?endpoint=http%3A%2F%2Fbigquery%3A9050&disable_auth=true
```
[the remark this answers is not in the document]
mii9000: removed `endpoint`

pkg/driver/bigquery/bigquery.go
```diff
@@ -0,0 +1,453 @@
...
+
+	// Execute the operation function with the client.
+	result, err := operation(ctx, client)
+
+	if err != nil {
+		return result, err
+	}
+
+	return result, nil
+}
```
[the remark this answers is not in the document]
mii9000: removed the `bigquery_client.go`, refactored the original code to reflect removal of `credentials` so it's not a copy anymore.

pkg/driver/bigquery/bigquery_client.go
```diff
@@ -0,0 +1,84 @@
...
+	opts := []option.ClientOption{option.WithScopes(config.scopes...)}
+	if config.endpoint != "" {
+		opts = append(opts, option.WithEndpoint(config.endpoint))
+	}
+	if config.disableAuth {
+		opts = append(opts, option.WithoutAuthentication())
+	}
+	if config.credentialFile != "" {
+		opts = append(opts, option.WithCredentialsFile(config.credentialFile))
+	}
```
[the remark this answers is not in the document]
mii9000: removed

pkg/driver/bigquery/bigquery.go
```diff
@@ -0,0 +1,525 @@
...
+
+	ctx := context.Background()
+	client, err := getClient(ctx, drv.config)
+	if err != nil {
+		return nil, err
+	}
+
+	drv.client = client
+	drv.context = &ctx
+
```
dossy, line 276: Now that you've incorporated the previous feedback, this is actually unnecessary.

`database/sql` provides a mechanism to get the underlying connection, through [func (*DB) Conn](https://pkg.go.dev/database/sql#DB.Conn).

If my understanding is correct, the underlying `go-gorm/bigquery` driver will give you back a `DB.Conn` that is this struct:

https://github.com/go-gorm/bigquery/blob/d4291d2dbfb0501df9f2221f625a34f6f881ef29/driver/connection.go#L10-L17

Inside it is the `*bigquery.Client` connection, etc.  We should use that and not our own private connection, so that all the connection management is handled by `database/sql`, including connection pooling, etc.

pkg/driver/bigquery/test_cred.json
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "service_account",
+  "project_id": "test-project",
+  "private_key_id": "someprivatekey",
+  "private_key": "\\n-----END PRIVATE KEY-----\n",
```
dossy, line 5: Should that first `\\n` be `\n`?

dossy (changes requested):

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,260 @@
...
+		"bigquery://bigquery:9050/test/dbmate_test?disable_auth=true",
+	}
+
+	for _, c := range cases {
+		t.Run(c, func(t *testing.T) {
+			_, err := url.Parse(c)
+			require.NoError(t, err)
+		})
+	}
+}
```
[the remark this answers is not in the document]
mii9000: You are right. Thanks for catching my oversight

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,260 @@
...
+	return sql.Open("bigquery", url)
+}
+
+func testBigqueryDriver(t *testing.T) *Driver {
+	u := dbutil.MustParseURL(getURL())
+	drv, err := dbmate.New(u).Driver()
+	require.NoError(t, err)
+
+	_, err = drv.Open()
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
mii9000: Thanks for the explanation. I will make the changes.

pkg/driver/bigquery/bigquery.go
```diff
@@ -0,0 +1,525 @@
...
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		_, err := createTable(*drv.context, drv.client)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
mii9000: copy that. no closures. either inlining or just reusable functions

pkg/driver/bigquery/bigquery.go
```diff
@@ -0,0 +1,525 @@
...
+type Driver struct {
+	migrationsTableName string
+	datasetID           string
+	projectID           string
+	log                 io.Writer
+	databaseURL         string
+	client              *bigquery.Client
+	context             *context.Context
+	config              *bigQueryConfig
+	sqlConnectionURL    string
```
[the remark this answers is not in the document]
mii9000: copy that

pkg/driver/bigquery/test_cred.json
mii9000: now that we are don't have a `bigquery_client.go` we don't really need this and much more simplified. will remove this.

pkg/driver/bigquery/bigquery.go
mii9000: Unfortunately, I was not able to find any way to get the underlying `bigQueryConnection` which has the `client` field. I tried a bit with reflection but was unsuccessful. `bigQueryConnection` is not exported from the gorm package as well. I am not an expert in Go but if you have any suggestions, I am all ears. Thanks.

mii9000:
@dossy Thanks a lot for your patience and reviews. I hope I was able to address all of the feedbacks.

pkg/driver/bigquery/bigquery.go
dossy: Using `*sql.Conn`'s [Raw()](https://pkg.go.dev/database/sql#Conn.Raw) gets us a handle to the underlying `bigQueryConnection`, but unfortunately the `go-gorm/bigquery` driver didn't make the `*bigquery.Client` struct member public. 😢

On top of that oversight, for some reason, they implemented `GetDataset()` and a `GetContext()` but no `GetClient()` -- they stopped just short of success.

I know there's all kinds of warnings around digging into structures and accessing unexported fields with reflection, but this might be the way to go, here.  Look at this test:

```
func TestGetClient(t *testing.T) {
	drv := testBigQueryDriver(t)

	db, err := drv.Open()
	require.NoError(t, err)
	defer dbutil.MustClose(db)

	conn, err := db.Conn(context.Background())
	require.NoError(t, err)
	defer conn.Close()

	err = conn.Raw(func(driverConn any) error {
		value := reflect.ValueOf(driverConn).Elem().FieldByName("client")
		value = reflect.NewAt(value.Type(), unsafe.Pointer(value.UnsafeAddr())).Elem()

		client := value.Interface().(*bigquery.Client)
		require.Equal(t, "test", client.Project())

		return nil
	})
	require.NoError(t, err)
}
```

Does that help?

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/amacneil/dbmate/v2/pkg/dbmate"
+	"github.com/amacneil/dbmate/v2/pkg/dbutil"
+)
+
+func getURL() string {
+	return os.Getenv("BIGQUERY_TEST_URL")
+}
+
+func testBigQueryDriver(t *testing.T) *Driver {
+	u := dbutil.MustParseURL(getURL())
```
dossy, line 20: `getURL()` is now only called from one place, so just inline it.

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	// connect database
+	db, err := drv.Open()
+	require.NoError(t, err)
+
+	return db
+}
+
+func TestGetDriver(t *testing.T) {
+	db := dbmate.New(dbutil.MustParseURL("bigquery://test/db_mate"))
```
dossy, line 46: For correctness, the URL should be `bigquery:///test/db_mate`.  Although, I suspect most people will make this same mistake, so we should also support URLs that are mis-constructed, but in our code and any documentation or examples, we should make sure to use the proper form.

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func TestConnectionString(t *testing.T) {
+	cases := []struct {
+		input    string
+		expected string
+	}{
+		{"bigquery://projectid/dataset", "bigquery://projectid/dataset?disable_auth=false"},
+		{"bigquery://projectid/location/dataset", "bigquery://projectid/location/dataset?disable_auth=false"},
+		{"bigquery://projectid/dataset?endpoint=http%3A%2F%2F0.0.0.0%3A9050", "bigquery://projectid/dataset?disable_auth=false"},
```
dossy, line 64: For this input, shouldn't the expected be `bigquery://projectid/dataset?endpoint=http%3A%2F%2F0.0.0.0%3A9050`?  No implied `disable_auth=false` as per my earlier comment.

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,252 @@
...
+	require.Equal(t, db.DatabaseURL.String(), drv.databaseURL.String())
+	require.Equal(t, "schema_migrations", drv.migrationsTableName)
+}
+
+func TestConnectionString(t *testing.T) {
+	cases := []struct {
+		input    string
+		expected string
+	}{
+		{"bigquery://projectid/dataset", "bigquery://projectid/dataset?disable_auth=false"},
```
dossy, line 62: Assuming that most users of dbmate will be using it with Google's actual BigQuery service and not the emulator, I don't think `disable_auth=false` should be the implied default.  For this input, I think `bigquery://projectid/dataset` should be the expected connection string.  What do you think?

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,252 @@
...
+		{"bigquery://projectid/location/dataset", "bigquery://projectid/location/dataset?disable_auth=false"},
```
dossy, line 63: What is this `/location` portion in the URL path?  Does the underlying `go-gorm/bigquery` driver understand this?

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,252 @@
...
+		{"bigquery://projectid/dataset?endpoint=http%3A%2F%2F0.0.0.0%3A9050", "bigquery://projectid/dataset?disable_auth=false"},
+		{"bigquery://projectid/location/dataset?endpoint=http%3A%2F%2F0.0.0.0%3A9050&disable_auth=true", "bigquery://projectid/location/dataset?disable_auth=true"},
```
dossy, line 65: Assuming we always alphabetize query parameters in connection strings we generate, shouldn't the expected be `bigquery://projectid/location/dataset?disable_auth=true&endpoint=http%3A%2F%2F0.0.0.0%3A9050`?  And, what is `/location` doing in the path, here?

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,252 @@
...
+		{"bigquery://bigquery:9050/test/dbmate_test?disable_auth=true", "bigquery://test/dbmate_test?disable_auth=true&endpoint=http://bigquery:9050"},
```
dossy, line 66: URL query strings should be properly URL-encoded, so the expected should be `bigquery://test/dbmate_test?disable_auth=true&endpoint=http%3A%2F%2Fbigquery%3A9050`.

dossy (changes requested):

pkg/driver/bigquery/bigquery_test.go
mii9000: https://github.com/go-gorm/bigquery/blob/master/driver/driver.go#L104-L106 - Yes it does understand `location`. Location path is to specify location for the BigQuery Dataset which is optional.

pkg/driver/bigquery/bigquery.go
mii9000: works like a charm!

pkg/driver/bigquery/bigquery_test.go
dossy: Maybe I'm mistaken, but I don't think that code does what you think it does.  Look at this example:

https://go.dev/play/p/hfmEYPPbhm9

```
func main() {
	u, _ := url.Parse("bigquery://projectid/location/dataset")
	fmt.Printf("u = %#v\n", u)

	fields := strings.Split(strings.TrimPrefix(u.Path, "/"), "/")
	fmt.Printf("len(fields) = %d\n", len(fields))
	for i, v := range fields {
		fmt.Printf("fields[%d] = %#v\n", i, v)
	}

	if len(fields) == 2 {
		fmt.Printf("config.location = fields[0] = %#v\n", fields[0])
	}
}
```

Notice how `config.location` will be set to `location` and not `dataset`.  That `go-gorm/bigquery` driver has a number of subtle bugs, because it accomodates improperly formed URLs, where they expect the project ID in the `host` portion of the URL, and the dataset ID in the `path` portion of the URL.

dossy:
@mii9000 Could you please append a section on BigQuery to the [Connecting to the Database](https://github.com/mii9000/dbmate/blob/baf2c8cdfda357749612f979192c84c59dc2a7be/README.md#L20-L25) section of the README, please?

Of particular importance is calling out the fact that the URL must encode the host and port of the endpoint they are connecting to, and if they want to connect to the default endpoint, the host portion of the URL must be empty, and caution the user to not put the project ID in the hostname portion of the URL.  Link to the Google document on [Application Default Credentials](https://cloud.google.com/docs/authentication/provide-credentials-adc) and to use the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to specify the path of the JSON file that contains their credentials, and document the optional query parameter and value `disable_auth=true` to disable authentication.

pkg/driver/bigquery/bigquery.go
```diff
@@ -527,18 +682,55 @@ func tableExists(ctx context.Context, client *bigquery.Client, datasetID, tableN
...
 }
 
-func connectionString(url *url.URL) string {
-	c, _ := configFromURI(url)
+func connectionString(u *url.URL) string {
+	var params, locationPath string
+
+	fields := strings.Split(strings.TrimPrefix(u.Path, "/"), "/")
+	projectID := u.Hostname()
+	dataset := fields[len(fields)-1]
```
dossy, line 690: The `projectID` should be `fields[0]`, not the `u.Hostname` for dbmate, and `dataset` should be `fields[1]`, therefore `len(fields) == 2` must be true, otherwise we should return an error -- but, unfortunately we would have to adjust a bit of tode to change the return signature from just `(string)` to `(string, error)` -- but, maybe it would be worth doing since `connectionString()` is called from only one place in the driver code, and only one place in the tests?

README.md
```diff
@@ -287,6 +288,28 @@ DATABASE_URL="clickhouse://username:password@127.0.0.1:9000/database_name?on_clu
 
 [See other supported connection options](https://github.com/ClickHouse/clickhouse-go#dsn).
 
+#### BigQuery
+Follow the following format for `DATABASE_URL` when connecting to actual BigQuery in GCP:
+
+```
+bigquery://projectid/location/dataset?disable_auth=true
+```
```
dossy, line 296: When connecting to actual BigQuery in GCP, `disable_auth=true` will never work, will it?

And, we are going to correctly express the project ID and dataset name in the URL path, so it should be:

```
bigquery:///projectid/dataset
```

README.md
```diff
@@ -287,6 +288,28 @@ DATABASE_URL="clickhouse://username:password@127.0.0.1:9000/database_name?on_clu
...
+
+`disable_auth` (optional) - Pass `true` to skip Authentication, use only for testing or connecting to emulator.
+
+*NOTE: Follow [this doc](https://cloud.google.com/docs/authentication/provide-credentials-adc) on how to set `GOOGLE_APPLICATION_CREDENTIALS` environment variable for proper Authentication*
+
+Follow the following format if trying to connect to a custom endpoint e.g. [BigQuery Emulator](https://github.com/goccy/bigquery-emulator)
+
+```
+bigquery://host:port/projectid/location/dataset
+```
```
dossy, line 311: Should be:

```
bigquery://host:port/projectid/dataset
```

dossy (changes requested):

pkg/driver/bigquery/bigquery.go
```diff
@@ -66,30 +51,89 @@ func (drv *Driver) CreateDatabase() error {
...
+		client := getClient(driverConn)
+		dataset := getDataset(driverConn)
+		err := client.Dataset(dataset).Create(ctx, &bigquery.DatasetMetadata{})
 		if err != nil {
 			return err
 		}
+		return nil
+	})
+
+	wg.Wait()
```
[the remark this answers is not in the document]
mii9000: My main point of concern was `con.Raw()` callback because from my limited knowledge of Golang, callbacks would not be sync.
image

README.md
mii9000: It will: https://github.com/go-gorm/bigquery/blob/master/driver/driver.go#L45-L47
Now that we are using the underlying client from the driver and not creating a new client within dbmate

README.md
mii9000: Since the underlying driver supports `location` should dbmate support location path as well?

pkg/driver/bigquery/bigquery.go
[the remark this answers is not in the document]
mii9000: But I did run the test file without `WaitGroup` once or twice and it ran fine. But when debugging, sometimes I noticed, the flow would go directly to the last statement of the function and then go into the callback

pkg/driver/bigquery/bigquery.go
mii9000: i have changed the logic a bit to use `net/url` but the tests remain the same so it should work as expected

pkg/driver/bigquery/bigquery.go
```diff
@@ -100,8 +144,13 @@ func (drv *Driver) CreateMigrationsTable(*sql.DB) error {
 		if err != nil {
 			return err
 		}
-
 		return nil
+	})
```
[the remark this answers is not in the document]
mii9000: done

pkg/driver/bigquery/bigquery.go
```diff
@@ -527,18 +682,55 @@ func tableExists(ctx context.Context, client *bigquery.Client, datasetID, tableN
...
+			locationPath = "/" + fields[1]
+		}
+	} else {
+		if len(fields) == 2 {
+			locationPath = "/" + fields[0]
+		}
 	}
 
-	return fmt.Sprintf("bigquery://%s%s/%s?%s", c.projectID, locationParam, c.dataSet, params)
+	return fmt.Sprintf("bigquery://%s%s/%s%s", projectID, locationPath, dataset, params)
```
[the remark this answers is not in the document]
mii9000: changed

pkg/driver/bigquery/bigquery.go
dossy: Once we get through the current batch of changes, I'll provide additional tests for `connectionString()`, as there's a lot of cases that aren't being tested.

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// drop any existing database
+	err := drv.DropDatabase()
+	require.NoError(t, err)
+
+	// ping database
+	err = drv.Ping()
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "dataset dbmate_test is not found")
+}
```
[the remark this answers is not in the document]
mii9000: The emulator does not implement INFORMATION tables so tests will fail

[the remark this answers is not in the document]
mii9000: The best I can do though is unit test it but that will not follow the same structure as other driver tests

README.md
mii9000: You are right. It's not able to do that. I will make the necessary changes for it.

pkg/driver/bigquery/bigquery.go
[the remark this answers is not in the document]
mii9000: removed all `WaitGroup`s

README.md
mii9000: added success test for success `Ping()` and `MigrationsTableExists()`

mii9000: the emulator does not support INFORMATION tables so not sure how I can follow the pattern like other driver tests for bq

dossy: > added success test for success `Ping()` and `MigrationsTableExists()`

Okay, now, temporarily change `BIGQUERY_TEST_URL` to `bigquery://bigquery:9050/test/location/dbmate_test?disable_auth=true` in your `docker-compose.yml` (no need to check this in) and run the tests.

mii9000: made the change and ran the tests. all tests passed. recorder: https://asciinema.org/a/T6K53MgHb6I89Lm3tFz58tOqN

dossy: > the emulator does not support INFORMATION tables so not sure how I can follow the pattern like other driver tests for bq

We can implement special tests that will only run against an actual Google BigQuery database URL conditionally, but we have to be very careful as the tests create and drop the dataset, which if someone runs the tests against an actual dataset in use, could be disasterous.

I'm talking about something like this:

```go
func TestGoogleBigQueryDumpSchema(t *testing.T) {
	t.Run("default migrations table", func(t *testing.T) {
		drv := testBigQueryDriver(t)

		if drv.databaseURL.Host != "" {
			t.Skipf("skipping test, BIGQUERY_TEST_URL pointing at %s and not at Google", drv.databaseURL.Host)
		}

		// test code continues here
	})
}
```

pkg/driver/bigquery/bigquery_test.go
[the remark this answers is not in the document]
dossy: Okay, so like I mentioned in [this comment](https://github.com/amacneil/dbmate/pull/523#discussion_r1526779349), we may need to implement a few special tests that will run against the actual Google BigQuery service and not the emulator.

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -288,3 +288,33 @@ func TestBigQueryPing(t *testing.T) {
...
+
+	db := prepTestBigQueryDB(t)
+	defer dbutil.MustClose(db)
+
+	// ping database
+	err := drv.Ping()
+	require.NoError(t, err)
+}
+
+func TestPostgresMigrationsTableExists(t *testing.T) {
```
dossy, line 307: fix: `Postgres` -> `BigQuery`

dossy (changes requested):

README.md
mii9000: I’m not sure I understand how these specific test cases would work in CI? If the test url connects to an actual BQ instance then it will pass/fail otherwise pass?

mii9000: @dossy i didn’t understand your comment about dbmate accepting host or host:port. That is how it’s working at

dossy: > I’m not sure I understand how these specific test cases would work in CI? If the test url connects to an actual BQ instance then it will pass/fail otherwise skip?

In CI, only test cases that can run against the emulator will be run; the tests that require the real Google BigQuery service should be skipped.

This way, we can implement tests that require the real Google BigQuery service to ensure that we're testing all the necessary functionality, but not put undue burden on anyone who wants to run the tests without having to set up a real GCP account, etc.

It makes accepting contribution for those drivers a little more difficult, but the alternatives are either not supporting drivers that have this requirement which hurts everyone, or not testing functionality that can only be tested against the real thing, which makes it hard to trust that dbmate will do the right thing.

> It might also be feasible to run the special cases against a free BQ instance? https://cloud.google.com/bigquery?hl=en#pricing

It might be possible to set up a free-tier GCP account with properly defined IAM permissions on a Service Account that could be used for testing in CI.  @amacneil, would you be willing to add a secret to the GitHub repo with a test GCP account's credentials JSON for use in CI test runs?

dossy: > i didn’t understand your comment about dbmate accepting host or host:port. That is how it’s working atm

I think it was a mistake for me to suggest that we support encoding the host and port in the connection string URL directly.

The [BigQuery Service Endpoint](https://cloud.google.com/bigquery/docs/reference/rest#service-endpoint) is `https://bigquery.googleapis.com`, so technically one could reasonably use the following connection string URL:

`bigquery://project/location/dataset?endpoint=https%3A%2F%2Fbigquery.googleapis.com`

The key issue here is being able to specify the correct scheme to use for the endpoint (e.g., `http` vs `https`), which if we try to infer the endpoint URL from the hostname and port portions of the connection string URL, we would not be able to infer the correct scheme for the endpoint.  We could try to assume that 443 means `https`, but does, say, port `1234` require `http` or `https`?  Can't know for sure, and there's no obvious way to encode it into the connection string.

In other words: now that we can get the project ID, location, and dataset from the underlying driver, we should probably just pass through the connection string URL as-is through to the underlying `go-gorm/bigquery` driver, let it do its connection string URL parsing, including encoding the endpoint as a query parameter, and read the values from the underlying `config` structure.

I'll push up a few commits that will hopefully clarify all of this.

pkg/driver/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,379 @@
...
+		require.NoError(t, err)
+		require.Contains(t, string(schema), fmt.Sprintf("CREATE TABLE `%s.%s.schema_migrations`", config.projectID, config.dataSet))
+		require.Contains(t, string(schema), "\n--\n"+
+			"-- Dbmate schema migrations\n"+
+			"--\n\n"+
+			"INSERT INTO schema_migrations (version) VALUES\n"+
+			"    ('abc1'),\n"+
+			"    ('abc2');\n")
+	})
+}
```
dossy, line 379: @mii9000 Try this test out and let me know what you think.

Authenticate with a service account that has the necessary IAM roles (see this [README](...) for details) and put a copy of the credentials JSON in `fixtures/bigquery/credentials.json` and give this a try:

```
$ make test \
  GOOGLE_APPLICATION_CREDENTIALS=/src/fixtures/bigquery/credentials.json \
 
 GOOGLE_BIGQUERY_TEST_URL=bigquery://$PROJECT_ID/$LOCATION/$DATASET
```

Replace `$PROJECT_ID/$LOCATION/$DATASET` with the project ID, region/location, and a test dataset name (remember, the dbmate tests will drop and create the dataset, so don't run tests against a dataset with data you can't afford to lose).

mii9000: image
@dossy All tests passed! Would it be better if the tests against the actual big query was in a separate file?

mii9000: So what else is required to get this PR merged @dossy?

dossy: @mii9000 Our test coverage against the actual BigQuery is pretty light, so I'm not especially confident that dbmate will work as expected through actual user interactions.  Users do all kinds of things that we don't have automated tests for.

It would be helpful if anyone who plans to actually use dbmate with BigQuery could pick up a dev build of it and actually start using it, and give feedback before it's merged -- there's no clear way for us to indicate that a particular driver may not be production-ready, yet.

I'm most concerned of dbmate dropping a user's entire dataset in production, when they didn't expect it to happen, or something terribly unfortunate like that.

mii9000: But the addition of bq driver is same as others. Other drivers are also dropping a database. Why bq should be any different? If its a concern then remove the drop database functionality?

mii9000: All the tests are working against the emulator. The only one missing is the dump schema one which can be covered with unit testing at least without requiring actual bq

pkg/driver/bigquery/bigquery_test.go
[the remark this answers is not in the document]
mii9000: we already did that

pkg/driver/bigquery/bigquery_test.go
mii9000: @dossy I can share a credentials.json from a trial account I created with you that will help you run the tests as well. Let me know if that will suffice with enough testing.

dossy: @mii9000 I have my own service account JSON that I'm testing with, but I have no actual projects on BigQuery that would use dbmate. I'd like someone who intends to use this for real projects to give it a try and give feedback on whether it works for them or not.

mii9000: I contributed this PR because my team intends to use this in Production. But if we are waiting for people to actually use this in Production before merging this PR or any PR going forward then at that rate it will take forever to run PRs.

I don't think it's completely up to maintainers or owners to validate how this tool will be used end-to-end. This PR checked all the boxes i.e. automated and manual tests, code reviews, standards, etc. I don't find this as a valid reason to delay releasing a new version based on this PR quite frankly. @dossy @amacneil

dossy: @mii9000 Do you require a tagged release before your team can use this?  Can you start with a build from this PR branch, and exercise the functionality with real migrations, and see if there's anything that doesn't work as expected, before this PR gets merged?

That's all I was hoping for, one last sanity check with real migrations before merging this.

If @amacneil is comfortable merging this as it currently is, that's up to him.  I'm a little more cautious, that's just my personal preference.

amacneil:
Thank you for the contribution @mii9000, and thanks @dossy for the review!

It looks like this PR is in a good state, and all the tests are passing. It any issues are discovered they can be fixed in a follow up PR.

I'll merge this now - please do some final tests on main. I'll make a release in a few days.

amacneil (approved):

PR #530, opened 2024-03-14
Adds support for templates when creating a new migration

jprokay:
Inspired by [Sqitch Templates](https://sqitch.org/docs/manual/sqitch-add/#templates)

Concerns:
I set the default directory to `~/.config/dbmate/templates` similar to sqitch. Should we consider different default paths based on the OS?
There are potential issues with needing to escape characters passed in via the CLI as mentioned in the #439


amacneil:
Have you found storing templates in your home folder useful? I would expect something like that to live in your repo.

jprokay:
@amacneil that's a great question. I do think most of my templates end up in the repo eventually. It depends on if I am working with others. The home folder can be convenient as a generic place to hold templates that I would use across different projects.

fast-90:
Is this still planned to be upstreamed? I would love to use this in a stable version of dbmate.

amacneil:
No. Having templates stored outside of your code repo seems against the KISS approach of dbmate.

You can also do this outside of dbmate easily by copy/pasting files.

PR #537, opened 2024-04-09
fix: embedded fs on windows

Dissociable:
Fixes #536

Issue #536, opened 2024-04-09
Embedded FS on Windows


Dissociable:
**Description**
Basically, this is a bug only for Windows. `readMigrationsDir` sets the dir path via `filepath.Clean(dir)` which changes the `/` to `\` on Windows.
`fs.ReadDir` Is not able to handle that. changing `filepath.Clean(...)` to `path.Clean(...)` would simply fix it.

- Version: v2.14.0
- Database: Postgresql
- Operating System: Windows 10

**Steps To Reproduce**
```
//go:embed migrate/migrations/*
var EmbeddedMigrations embed.FS


	dbMate := dbmate.New(getAddr(dbName))
	dbMate.FS = ent.EmbeddedMigrations
	dbMate.MigrationsDir = []string{"migrate/migrations"}
	fmt.Println("Migrations:")
	migrations, err := dbMate.FindMigrations()
	if err != nil {
		panic(err)
	}
```

**Expected Behavior**
Find the embedded files at `migrate/migrations`

dossy:
@Dissociable thanks for reporting this.

This makes me wonder: is it worth investing time in extending our CI test to add a Windows runner to test dbmate on, as well?  Similarly, should we add a macOS runner, too?

I know that GitHub hosted Windows and Mac action runners are [considerably more expensive](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers), and I don't know if @amacneil wants to incur those costs for the project or not.

boludoz:
@Dissociable contact with me in discord

PR #545, opened 2024-04-21
Revert "fix: embedded fs on windows (#537)"

amacneil:
I am not going to merge this PR. The point of it is to prove that the issue described in https://github.com/amacneil/dbmate/issues/536 and fixed by https://github.com/amacneil/dbmate/pull/537 is reproducible now that we run CI tests on Windows.

Reverts commit 716b623d095a384189f37e7dccc4f230e2786e89.

### DIFF
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -8,7 +8,6 @@ import (
 	"io/fs"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


amacneil:
As expected, CI failed on Windows and passed on all other platforms.

PR #555, opened 2024-06-17
Spanner Postgres interface support

grahamhoyes:
Adds support for the [GCP Spanner Postgres Interface](https://cloud.google.com/spanner/docs/postgresql-interface), which only requires a slight modification to the postgres driver.

This will not work with a GoogleSQL-flavored Spanner database; support for that will require a dedicated driver.

The new functionality is accessed with a custom `spanner-postgres://` scheme. This is rather un-standard and something I just made up, so the exact scheme is open for debate. We shouldn't use just `spanner://`, since that may cause confusion down the line when a driver for Spanner with the GoogleSQL dialect is added.

Related discussion: https://github.com/amacneil/dbmate/discussions/369

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,32 @@ bigquery://host:port/projectid/location/dataset?disable_auth=true
 
 `disable_auth` (optional) - Pass `true` to skip Authentication, use only for testing and connecting to emulator.
 
+#### Spanner (PostgreSQL Interface)
+
+Spanner support is currently limited to databases using the [PostgreSQL Dialect](https://cloud.google.com/spanner/docs/postgresql-interface), which must be chosen during database creation. For future Spanner with GoogleSQL support, see [this discussion](https://github.com/amacneil/dbmate/discussions/369).
+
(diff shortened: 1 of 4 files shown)


amacneil:
Great idea, @dossy care to review?

dossy (approved):
I don't have a Google Spanner instance handy to test this against, but static analysis review of the code LGTM.

grahamhoyes:
> I don't have a Google Spanner instance handy to test this against, but static analysis review of the code LGTM.

If anyone else would like to play around with it, Google claims that their emulator (also used in the tests here) is feature-compatible with cloud spanner:

```sh
docker run -p 5432:5432 gcr.io/cloud-spanner-pg-adapter/pgadapter-emulator
```

Whatever database you attempt to connect to will automatically be created. No username/password required.

grahamhoyes:
Gave this a more thorough test on a real Spanner instance today, looks to be behaving as expected.

Anything I can help with to get this merged for the next release?

dossy:
@grahamhoyes Thanks for following up.  No one has raised any concerns, so I'll go ahead and merge now.

PR #573, opened 2024-08-20
add go install as installation option

sombriks:
use go install results in a functional dbmate installation, as far as my tests goes.

please add this as installation option in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ $ scoop install dbmate
 
 **GO Install**
 
-Remember to add `~/go/bin` into your `$PATH` if you decide to install dbmate this way.
+Remember to add `"$(go env GOPATH)/bin"` into your `$PATH` if you decide to install dbmate this way.
 
 ```sh
 go install github.com/amacneil/dbmate/v2
--- a/README.md
+++ b/README.md
@@ -89,6 +89,14 @@ Install using [Scoop](https://scoop.sh)
 $ scoop install dbmate
 ```
 
+**GO Install**
+
+Remember to add `~/go/bin` into your `$PATH` if you decide to install dbmate this way.
+
+```sh
+go install github.com/amacneil/dbmate/v2
+```
+
 **Docker**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,6 +89,14 @@ Install using [Scoop](https://scoop.sh)
 $ scoop install dbmate
 ```
 
+**GO Install**
+
+Remember to add `~/go/bin` into your `$PATH` if you decide to install dbmate this way.
```
dossy, line 94: `~/go/bin` is not correct in all cases.  A more generally correct recommendation would be `$(go env GOPATH)/bin`.

amacneil:
I don't recommend this install option:

* No guarantee that go bin is in your $PATH
* Installing "latest commit on main" is never a good idea, you want a released version
* if you happen to run this inside an existing module directory you get

```sh
$ go install github.com/amacneil/dbmate/v2
no required module provides package github.com/amacneil/dbmate/v2; to add it:
	go get github.com/amacneil/dbmate/v2
```

Running the proposed `go get` command adds a ton of unrelated dependencies to your project.

Generally I don't see any reason you would want to build dbmate from source unless you are hacking on it or using as a library. Those that are doing so should already know the appropriate commands to run.

sombriks:
> I don't recommend this install option:
> 
> * No guarantee that go bin is in your $PATH
> * Installing "latest commit on main" is never a good idea, you want a released version
> * if you happen to run this inside an existing module directory you get
> 
> ```shell
> $ go install github.com/amacneil/dbmate/v2
> no required module provides package github.com/amacneil/dbmate/v2; to add it:
> 	go get github.com/amacneil/dbmate/v2
> ```
> 
> Running the proposed `go get` command adds a ton of unrelated dependencies to your project.
> 
> Generally I don't see any reason you would want to build dbmate from source unless you are hacking on it or using as a library. Those that are doing so should already know the appropriate commands to run.

i had no idea it was intentional and understand the reasons. feel free to close this PR.

amacneil:
Also I should mention: Since we import sqlite via cgo, you will need a C toolchain installed to compile that. So ya, I highly recommend sticking to prebuilt binaries.

PR #581, opened 2024-09-20
Turso support

eynopv:
This pull request adds basic support for [Turso](https://turso.tech/).
Current implementation does not include database creation and deletion as connection to Turso happens with their provided urls in following format:
`libsql://[databaseName]-[organizationName].turso.io`
It would have been nice to parse the url by `-`, but both `databaseName` and `organizationName` can contain `-` character.
To support database creation and drop, the format for `DATABASE_URL` can be changed to `turso://[organizationName]/[databaseName]` and then build the connection string from that, if it's something desired.
It's also worth noting that database can belong to a [group](https://docs.turso.tech/cli/db/create#create-database-in-a-specific-group) and could be added to the `DATABASE_URL`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ For a comparison between dbmate and other popular database schema migration tool
     - [ClickHouse](#clickhouse)
     - [BigQuery](#bigquery)
     - [Spanner](#spanner)
+    - [Turso](#turso)
   - [Creating Migrations](#creating-migrations)
   - [Running Migrations](#running-migrations)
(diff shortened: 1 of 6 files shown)


peteraba:
@amacneil this would be a very welcome addition.

eynopv:
I’m happy to finish this PR. I just need clarification on what needs to be done for this to go forward and for the contribution to be accepted.

kalbasit:
@amacneil any chance you could give this a review? I'm eager to use it as well...

Issue #237, opened 2021-09-10
sqlite fts5 migration 


ak5:
I want to make a table with the fts5 sqlite extension like so:

```sql
CREATE VIRTUAL TABLE mytable_fts using fts5(content="mytable", mycolumn);
```

this works in sqlite3 but not in a dbmate migration with the error `Error: no such module: fts5`.

amacneil:
We currently build sqlite into the dbmate binary, with extension support disabled:

https://github.com/amacneil/dbmate/blob/main/Makefile#L6

I believe this was necessary because we want to build a static binary (that can work in alpine containers, etc). You could try changing those build flags and rebuilding sqlite with fts5 enabled?

I don't know much about how sqlite extensions work. Do they always need to be compiled into the binary or can they be loaded at runtime?

Open to suggestions on how to improve the current state of things. Maybe we could use the system sqlite executable if one is available, and fall back to our bundled version.

ak5:
I think using system sqlite binaries as a flag like `--sqlite-binary=/path/to/sqlite3` is probably the best/easiest to implement solution without breaking anything...

ak5:
For now I am going with making my own binary, seems like I can handle that, thanks!

amacneil:
It also might be fine to just bundle common extensions?

I’m not exactly sure how to use the system sqlite, but we could look into it.

ak5:
Maybe just do what ubuntu or another common distro does in that case?

Enrico204:
Using system sqlite3 means that the binary is no more a static binary that can runs "everywhere" (meaning that it's not tied to a particular version of sqlite or userland in general).

One possibility is to build two flavors: the main one like we are doing right now, and a different one that uses libraries from the O.S. I don't know if it's possible and how to do this, however this can be a solution that doesn't require too much effort (I think).

PR #582, opened 2024-10-05
Enable FTS5 in SQLite migrations

jadudm:
This statically compiles in the FTS5 module, allowing dbmate to handle migrations that use the full-text search features of SQLite.

## History

Issue #237 raised the question of handling migrations that included FTS5 tables. It was later converted to a discussion (#357).

## Solution

It is possible to enable the feature in the statically-compiled SQLite that ships with dbmate via a feature flag. Adding the `sqlite_fts5` flag to the `Makefile` embeds this feature statically.

## Testing

An additional test was added with a `CREATE VIRTUAL TABLE` statement that leverages the FTS5 module. This test fails when dbmate is compiled _without_ the feature flag. After adding the flag, this test passes.

`make test` and `make docker-all` pass.

Confirmed that it works in practice locally on Ubuntu 22.04.  

## Additional info

For information about the extension:
https://www.sqlite.org/fts5.html

For the feature flag for mattn/go-sqlite3:
https://github.com/mattn/go-sqlite3/?tab=readme-ov-file#feature--extension-list

### DIFF
--- a/pkg/driver/sqlite/sqlite_test.go
+++ b/pkg/driver/sqlite/sqlite_test.go
@@ -406,17 +406,11 @@ func TestSQLiteQuotedMigrationsTableName(t *testing.T) {
 	})
 }
 
-func TestSQLiteFTS5MigrationsTable(t *testing.T) {
-	drv := testSQLiteDriver(t)
-	drv.migrationsTableName = "test_migrations"
-
-	// prepare database
+func TestSQLiteFTS5Available(t *testing.T) {
 	db := prepTestSQLiteDB(t)
 	defer dbutil.MustClose(db)
-	err := drv.CreateMigrationsTable(db)
-	require.NoError(t, err)
 
-	// this only passes if the FTS5 module is staticall compiled in to the SQLite driver
-	_, err = db.Exec("CREATE VIRTUAL TABLE a USING fts5(b, c)")
+	// this only passes if the FTS5 module is statically compiled in to the SQLite driver
+	_, err := db.Exec("CREATE VIRTUAL TABLE a USING fts5(b, c)")
 	require.NoError(t, err)
 }
--- a/pkg/driver/sqlite/sqlite_test.go
+++ b/pkg/driver/sqlite/sqlite_test.go
@@ -405,3 +405,18 @@ func TestSQLiteQuotedMigrationsTableName(t *testing.T) {
 		require.Equal(t, `"fooMigrations"`, name)
 	})
 }
+
+func TestSQLiteFTS5MigrationsTable(t *testing.T) {
+	drv := testSQLiteDriver(t)
+	drv.migrationsTableName = "test_migrations"
+
+	// prepare database
+	db := prepTestSQLiteDB(t)
+	defer dbutil.MustClose(db)
+	err := drv.CreateMigrationsTable(db)
+	require.NoError(t, err)
+
+	// this only passes if the FTS5 module is staticall compiled in to the SQLite driver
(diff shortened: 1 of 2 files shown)


pkg/driver/sqlite/sqlite_test.go
```diff
@@ -405,3 +405,18 @@ func TestSQLiteQuotedMigrationsTableName(t *testing.T) {
...
+
+func TestSQLiteFTS5MigrationsTable(t *testing.T) {
+	drv := testSQLiteDriver(t)
+	drv.migrationsTableName = "test_migrations"
+
+	// prepare database
+	db := prepTestSQLiteDB(t)
+	defer dbutil.MustClose(db)
+	err := drv.CreateMigrationsTable(db)
+	require.NoError(t, err)
```
dossy, line 417: There's no need to repeat testing `drv.CreateMigrationsTable()` in this test, as it's already covered by other tests.

This should be sufficient to replace L410-L417:

```
		drv := testSQLiteDriver(t)
		db := prepTestSQLiteDB(t)
		defer dbutil.MustClose(db)
```

pkg/driver/sqlite/sqlite_test.go
```diff
@@ -405,3 +405,18 @@ func TestSQLiteQuotedMigrationsTableName(t *testing.T) {
 		require.Equal(t, `"fooMigrations"`, name)
 	})
 }
+
+func TestSQLiteFTS5MigrationsTable(t *testing.T) {
```
dossy, line 409: As this test really shouldn't duplicate testing migration functionality, the name should be updated to reflect that.

I propose `TestSQLiteFTS5Available` as the test name, as that is what it is testing for.

pkg/driver/sqlite/sqlite_test.go
```diff
@@ -405,3 +405,18 @@ func TestSQLiteQuotedMigrationsTableName(t *testing.T) {
...
+	drv := testSQLiteDriver(t)
+	drv.migrationsTableName = "test_migrations"
+
+	// prepare database
+	db := prepTestSQLiteDB(t)
+	defer dbutil.MustClose(db)
+	err := drv.CreateMigrationsTable(db)
+	require.NoError(t, err)
+
+	// this only passes if the FTS5 module is staticall compiled in to the SQLite driver
```
dossy, line 419: nit: `staticall` -> `statically`

dossy (changes requested):
Please let me know if I need to elaborate on any of the review feedback.  Thanks!

pkg/driver/sqlite/sqlite_test.go
jadudm: Based on reading `prepTestSQLiteDB`, I think that `drv := testSqliteDriver(t)` isn't needed here? Either way, understood, simplified, agreed.

pkg/driver/sqlite/sqlite_test.go
jadudm: Agreed. Renamed.

pkg/driver/sqlite/sqlite_test.go
jadudm: Nits always accepted. Fixed.

jadudm:
No elaboration needed at all; many thanks for the review!

dossy (approved):

valeriakononenko:
Hi! When do you plan to release it?

amacneil:
Done!

https://github.com/amacneil/dbmate/releases/tag/v2.22.0

PR #589, opened 2024-11-07
feat: support multiple queries for clickhouse using up/down-sections

dsagoyaneleven:
Currently, ClickHouse does not support executing multiple queries. However with these changes, it becomes possible.
We just put each separate up- or down-query in sections, but sections are optional.

### DIFF
--- a/pkg/dbmate/db.go
+++ b/pkg/dbmate/db.go
@@ -385,12 +385,17 @@ func (db *DB) Migrate() error {
 		}
 
 		execMigration := func(tx dbutil.Transaction) error {
-			// run actual migration
-			result, err := tx.Exec(parsed.Up)
-			if err != nil {
-				return drv.QueryError(parsed.Up, err)
-			} else if db.Verbose {
-				db.printVerbose(result)
+			if len(parsed.UpSections) != 0 {
+				for _, section := range parsed.UpSections {
+					if err = db.execQuery(drv, tx, section); err != nil {
+						return err
+					}
+				}
+			} else {
+				// run actual migration
+				if err = db.execQuery(drv, tx, parsed.Up); err != nil {
+					return err
+				}
 			}
 
 			// record migration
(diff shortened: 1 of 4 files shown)


amacneil:
This seems like a nice solution that avoids full sql parsing.

Imo we could avoid the separate `-- migrate:up-section` and just allow multiple `-- migrate:up` blocks?

Curious to hear what others think.

dossy:
> Imo we could avoid the separate `-- migrate:up-section` and just allow multiple `-- migrate:up` blocks?

+1 on allowing multiple `-- migrate:up` (and, similarly, `-- migrate:down`) blocks.

jakobo:
> Curious to hear what others think.

Chiming in as a current goose user who couldn't use `dbmate` earlier because of the multiple clickhouse queries issue.

Multiple `--migrate:up` and `--migrate:down` blocks, each with their own settings, is perfectly reasonable. It's also IMO the simplest to explain: you can break multiple statements up by adding additional migrate directives to your file. It's your responsibility to manage a partial migration in this situation, and only recommended when it's unreasonable to treat these actions each as their own migration.

For me, It turns seven clickhouse migrations into a single file and has zero impact on the postgres migration files.

peick (approved):

dossy (changes requested):
Rework the PR to not introduce `migrate:up-section` and `migrate:down-section`, and instead change `ParsedMigration`'s `Up` from `string` to `[]string` and `UpOptions` from `ParsedMigrationOptions` to `[]ParsedMigrationOptions`, and similar for `Down` and `DownOptions`.

The tests are lacking cases that demonstrate the ability to specify different migration options per each up or down block.

The PR is also missing updates to the documentation that document this feature, as well.

dossy:
Closing this PR in favor of #620, which allows specifying multiple `-- migrate:up` and `-- migrate:down` sections .

PR #614, opened 2025-01-22
Remove binary_parameters from pg_dump database url

ultrotter:
These are sometimes needed in the pq library, but are not supported when calling pg_dump.


amacneil (approved):
never heard of it but you are correct, seems to be pq specific

dossy:
Interesting.

https://github.com/lib/pq/issues/435

PR #620, opened 2025-02-23
feat: add multistatement separator

igorvoltaic:


dossy:
If we support multiple up- and down-migrations in a single file, we don't need a `-- migrate:statement-end` - each statement is ended by either a subsequent `-- migrate:up` or `-- migrate:down` or EOF.

pkg/dbmate/migration.go
```diff
@@ -39,9 +40,9 @@ func (m *Migration) Parse() (*ParsedMigration, error) {
 
 // ParsedMigration contains the migration contents and options
 type ParsedMigration struct {
-	Up          string
+	Up          []string
 	UpOptions   ParsedMigrationOptions
```
dossy, line 44: I think `UpOptions` and `DownOptions` need to be turned into arrays, as well, as each migration can have its own options.

pkg/dbmate/migration.go
```diff
@@ -66,6 +67,7 @@ var (
 	whitespaceRegExp      = regexp.MustCompile(`\s+`)
 	optionSeparatorRegExp = regexp.MustCompile(`:`)
 	blockDirectiveRegExp  = regexp.MustCompile(`^--\s*migrate:(up|down)`)
+	separatorRegExp       = regexp.MustCompile(`(?m)^\s*--\s*migrate:separator\s*$`)
```
dossy, line 70: If we support multiple up- and down-migrations in a single file, we don't need a `-- migrate:statement-end` since each statement is already ended by either a subsequent `-- migrate:up` or `-- migrate:down` or EOF.

pkg/dbmate/migration.go
```diff
@@ -102,9 +104,9 @@ func parseMigrationContents(contents string) (*ParsedMigration, error) {
 	downBlock := substring(contents, downDirectiveStart, len(contents))
 
 	parsed := ParsedMigration{
-		Up:          upBlock,
+		Up:          splitSQLStatements(upBlock),
 		UpOptions:   parseMigrationOptions(upBlock),
-		Down:        downBlock,
+		Down:        splitSQLStatements(downBlock),
 		DownOptions: parseMigrationOptions(downBlock),
```
dossy, line 110: This approach isn't properly handling options on each migration.

Implementing this feature correctly will require a redesign of the `parseMigrationContents()` from the current naive approach to one that splits the file into regions and then parses each region, filling arrays as it goes.

dossy (changes requested):

pkg/dbmate/migration.go
igorvoltaic: I went a bit other way and implemented file sections

pkg/dbmate/migration.go
igorvoltaic: Thanks for the tip, I've used `-- migrate:up` as a separator to split file by migration sections and process each section with existing logic

pkg/dbmate/migration.go
igorvoltaic: should be fine by now

dossy: Could you add a test that would otherwise fail if the proper support is not implemented correctly?

dossy:
We already have tests that assert that each migration file must have a [`migrate:up`][1] and [`migrate:down`][2] and that [`migrate:up` must appear before `migrate:down`][3].

Please add tests for the following cases to make it clear what the expected behaviors are:

  * two up, one down:

  ```sql
  -- migrate:up
  select 1

  -- migrate:up
  select 2

  -- migrate:down
  select 3
 ```

  * one up, two down

  ```sql
  -- migrate:up
  select 1

  -- migrate:down
  select 2

  -- migrate:down
  select 3
 ```

Adding these tests would serve to clarify the expected behavior intended by this change, whether these should be accepted as valid inputs or rejected with an error.

[1]: https://github.com/amacneil/dbmate/pull/620/files#diff-b72af22c75eff5a7a050904af16e75a69aacdc9481296fdbf834335fd3589963R140-R147
[2]: https://github.com/amacneil/dbmate/pull/620/files#diff-b72af22c75eff5a7a050904af16e75a69aacdc9481296fdbf834335fd3589963R149-R156
[3]: https://github.com/amacneil/dbmate/pull/620/files#diff-b72af22c75eff5a7a050904af16e75a69aacdc9481296fdbf834335fd3589963R86-R95

pkg/dbmate/migration_test.go
```diff
@@ -88,16 +101,30 @@ create table users (id serial, name text);
...
-		require.Equal(t, "-- migrate:up transaction:false\nALTER TYPE colors ADD VALUE 'orange' AFTER 'red';\n", parsed.Up)
-		require.Equal(t, false, parsed.UpOptions.Transaction())
+		require.Equal(t, "-- migrate:up transaction:false\nALTER TYPE colors ADD VALUE 'green' AFTER 'red';\n", parsedSecondSection.Up)
+		require.Equal(t, false, parsedSecondSection.UpOptions.Transaction())
 
-		require.Equal(t, "-- migrate:down transaction:false\nALTER TYPE colors ADD VALUE 'orange' AFTER 'red';\n", parsed.Down)
-		require.Equal(t, false, parsed.DownOptions.Transaction())
+		require.Equal(t, "-- migrate:down transaction:false\nALTER TYPE colors ADD VALUE 'green' AFTER 'red';\n", parsedSecondSection.Down)
+		require.Equal(t, false, parsedSecondSection.DownOptions.Transaction())
 	})
```
dossy, line 128: I think this test would be more meaningful if:

  * the second pair of up/down had `transaction:true`
  * a third pair of up/down have no options at all, but within the same migration file
  * a fourth and fifth pair have mixed options, the fourth having options on the up but not on the down, and the fifth having options on the down but not the up

I think this would fully cover all the possible scenarios, and also demonstrate the options are actually being parsed independently in each section and not carried over or reused from a previous section.

dossy (changes requested):

dossy:
> * **Sequential Versioning**:
>   Migrations derive versions from the file's timestamp plus a section index (e.g., `20231010120000-0`, `20231010120000-1`), ensuring uniqueness.

I personally dislike this, and this breaks backwards-compatibility, as [the current behavior to only record the version number from the migration filename](https://github.com/amacneil/dbmate/blob/c1c1ae6385ce91d13faf991dbea6db239357dd49/README.md?plain=1#L365) is the expected behavior and clearly documented.

Migration files should either succeed or fail as a whole, and the version recorded should be for the migration file, not individual sections.  What purpose does recording individual sections serve?

igorvoltaic:
> What purpose does recording individual sections serve?

This question is more likely to arise once sections are fully implemented. I assumed users would desire section-specific rollbacks, this functionality could potentially be introduced as a configurable option via section props rather than relying solely on migration versions.

pkg/dbmate/migration.go
igorvoltaic: the implementation is the same as in main branch, no changes here

dossy:
> > What purpose does recording individual sections serve?
> 
> This question is more likely to arise once sections are fully implemented. I assumed users would desire section-specific rollbacks, this functionality could potentially be introduced as a configurable option via section props rather than relying solely on migration versions.

Perhaps.  Separating that out into a separate PR and not breaking backwards-compatibility in this one is my preference, anyway.

pkg/dbmate/version.go
```diff
@@ -1,4 +1,4 @@
 package dbmate
 
 // Version of dbmate
-const Version = "2.27.0"
+const Version = "2.28.0"
```
dossy, line 4: Bumping the version number is done as part of the release process in its own separate PR (e.g., #644) and not in a feature PR.  Can you revert this change?  Thanks.

dossy (changes requested):

dossy:
One final change required for this PR: updating the "Creating Migrations" section in the README to document this new functionality.

dossy (approved):
Thanks for updating the documentation.

The failing check appears to be unrelated to the changes in this PR, so I'm approving it and the failing check will be addressed in a separate PR.

dossy:
Once PR #660 lands, merging that into this PR should get the checks passing again.

dossy:
@igorvoltaic FYI, #660 was merged today.  Merge latest upstream `main` into this PR branch to pick up the change, and all checks should pass.

PR #628, opened 2025-03-12
Add dump column in schema_migration and sync command line argument

Sentenzius:
I've added 2 new features to dbmate that can be helpful for many.

1. I added a column in migration_schema named 'dump'.

This column contains the 'down' migration that need to be executed to rollback the applied migrations. This is useful in this scenario: 

- We have installed a software compiled in the year 2025.
- For some reason this software isn't stable, we've got to rollback to a software compiled in 2023.
- Both during install use dbmate.
- The older software even checking the schema_migrations table, wouldn't know how to rollback the migrations done in 2024 and 2025, because the 2023 installer doesn't have those .sql files.
- Storing the dump of the 'down' queries give older installers to correctly rollback future migrations.

2. I added a command line argument named 'sync'.

This command automatically execute all 'up' migrations available in one shot, or rollbacks future migrations found in existing database thanks to the 'dump' column added in schema_migrations.

P.S. In case the current schema_migrations doesn't have the 'dump' column it is added, and it would be empty at the beginning. But if the .sql files provided to dbmate contain the migration versions already applied, dbmate will also populate the 'dump' column accordingly. So it would work also with existing database. If this feature is not used, does not harm.

P.P.S. Works perfectly with MySQL, I tried also to adapt code of others languages, they compile all, but I don't have tested them.


dossy:
Applying down migrations is generally a dangerous procedure that results in irreversible data loss.  Making it _too_ easy to unintentionally perform them through a "synchronize" action feels especially dangerous.

The idea of letting an older version of software destroy data used by a newer version seems to me like a foot-gun of epic proportions.

If there are others who have actual use cases that require this functionality, they should speak up in favor of this change.  I do not, so I'll recuse myself.

Thanks for the thorough work on the PR, though.

Sentenzius:
Yes, the downgrade can expose to loss of data, but in our use case sometimes is necessary.

Even if the PR will be rejected your answer @dossy has been useful because I unserstood better how to use down migrations! 

Because our purpose of the migration is to adapt database to software changes. But we apply down migration for example because a stored procedure isn't compatible with previous versions.

Instead for example if a migration add a column, and in that column is present important data, a down migration shouldn't remove that column. It won't use it, but should delete that.

Previously instead I've always thought down migrations as the contrary of up migrations.

Sentenzius:
I've been using this forked version for a while (with possibility to rollback with sync command line argument) and works pretty good. 

Recently I'm adding the feature to dump in 'dump' column of schema migrations not only the down migration but both up and down migrations.

The purpose of this is to use another db as 'source' of migration data, as an alternative of migration files. In particular one can use --migration-dir and specify a folder for migrations to be applied or --migration-url and specify a database where to search for schema_migration table and retrieve migrations from 'dump' column of that table.

The aim is: I have 2 databases: 
- dbA with certain migration applied.
- dbB with other migrations (older or newer it doesn't matter).

We need to update structure of dbB to be the same schema of dbA. But we don't have the original migrations files, so we can rely only of the data present in the db. That's the reason behind this development.

@dossy @amacneil Consider again if you are interested in such feature, I would be glad if you want to integrate it in the main project (maybe a little more work is needed to test other dbs, I tested mainly mysql).

dossy:
> @dossy @amacneil Consider again if you are interested in such feature, I would be glad if you want to integrate it in the main project (maybe a little more work is needed to test other dbs, I tested mainly mysql).

Personally, I'm not, and I think the approach may be missing one of the key reasons for schema management tools to exist in the first place.

Version control and auditability are key aspects of schema management, and it's reasonably easy to do that with files in a filesystem, and not so easy with rows in a database table: few DBMS softwares are implemented with immutable journals, which would be the first must-have requirement for version control and auditability of the database containing the mutation scripts' contents.

If all you need is a SQL script runner against a target database, `dbmate` is practically overkill of a solution: a simple shell script that assembles the script to run and then uses a CLI tool to spool it into your target database is all you need.  Whether your shell script assembles that script from files in a filesystem, or the results of a database query, or any other data source, you just implement that in your script and you've got a perfectly fit-for-purpose solution for your problem.

Before today, suggesting people write the shell script might have been a technological challenge that not everyone has the skills to overcome, but all of the current AI code-generating agents can do this, so I'm not as apprehensive about making it as a serious suggestion.

Of course, this is all only my personal opinion, which I'm sharing because you asked me directly, and is only meant to help provide another point of view to the discussion.

Issue #217, opened 2021-06-13
Support Sqlite Pragmas in Migrations


ak5:
Currently with latest version (dbmate version 1.12.0) I cannot do:

```sql
PRAGMA journal_mode = WAL;
```

with the error:

```
Error: cannot change into wal mode from within a transaction
```

It would be nice to be able to rely on migrations to setup my databases completely without having to remember to change pragmas manually or via application code

amacneil:
It should work if you use the `migrate:up transaction:false` flag in your migration file.

PR #636, opened 2025-04-11
Mention sqlite transaction:false in README.md

malteneuss:
Add hint from https://github.com/amacneil/dbmate/issues/217 to Readme for sqlite to set common PRAGMA settings


dossy (approved):
Hope this helps others avoid confusion and frustration.  Thanks!

PR #641, opened 2025-04-18
Prefer `varchar` over `varchar(n)` on Postgres-like databases

nstylo:


dossy (approved):
Approved.

See also: https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default

PR #676, opened 2025-08-20
Fix clickhouse schema dump with non-default database

Flamefork:
This fixes the issue of using a generated `schema.sql` with ClickHouse in Docker.
The ClickHouse entrypoint runs every `.sql` file in `/docker-entrypoint-initdb.d` during server startup, but it executes them against the default database.

It works if the schema dump was created from the default database, but it fails otherwise:
```
/entrypoint.sh: running /docker-entrypoint-initdb.d/schema.sql
Received exception from server (version 25.3.6):
Code: 60. DB::Exception: Received from 127.0.0.1:9000. DB::Exception: Table default.schema_migrations does not exist. Maybe you meant my_database_name.schema_migrations?. (UNKNOWN_TABLE)
```
This fix prefixes `migrations_table_name` with `db_name.` in the generated schema, making it consistent with the table creation statements and with other drivers.

### DIFF
--- a/pkg/driver/clickhouse/clickhouse.go
+++ b/pkg/driver/clickhouse/clickhouse.go
@@ -141,8 +141,8 @@ func (drv *Driver) escapeString(str string) string {
 
 // CreateDatabase creates the specified database
 func (drv *Driver) CreateDatabase() error {
-	name := drv.databaseName()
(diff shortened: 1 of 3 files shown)


Flamefork:
@amacneil @sofuture @dossy could you please take a look? thanks!

PR #693, opened 2025-10-23
Add --set-role flag

Ragge-dev:
## Background
Hello :wave: 
We are using dbmate to handle migrations in our Postgresql databases (and we think it is great :slightly_smiling_face:). We have a role setup with an `admin` account which is the owner of all tables and then we want to have a separate `db_migration` role, which we login to the db with to run migrations. 

The problem we face is that with dbmate we always run the db migrations as the user we login as, running 
```
dbmate --url postgres://db_migration:@127.0.0.1:5432/?sslmode=disable migrate
```
any new tables would be created and be owned by the `db_migration` role. For different reasons we would prefer not to use the `admin` user to login but instead login as the `db_migration` user and run the migrations as the `admin` role. You could add `SET ROLE admin` into the migration files as well but that makes it easier to make a mistake and adds boilerplate.

## The Feature
Since we want to keep our postgresql admin role separate from the account running the database migrations we tested to add a new flag `--set-role` which allows a user to specify what role the database migrations should be run as. Using this we are able to run:
```
dbmate --url postgres://db_migration:@127.0.0.1:5432/?sslmode=disable --set-role admin migrate
```
And then, even if we use the `db_migration` role to auth towards postgres, we will run `SET ROLE admin` before any dbmate command. 
___
We are curious if you think such a feature would be a good addition to dbmate. Let me know your thoughts, if you think it fits in this or another form or if it does not match your vision for dbmate.

Best
Ragnar

---

> [!NOTE]
> **Medium Risk**
> Changes core migration execution flow and driver interface across all backends, and modifies Postgres connection behavior by injecting `SET ROLE` into multiple operations; issues would surface at runtime during migrations/DDL.
> 
> **Overview**
> Adds a new CLI/env option `--set-role` (`DBMATE_SET_ROLE`) that propagates a `DatabaseRole` setting into dbmate so database operations can be executed under a specified role.
> 
> Extends the driver interface with `PrepareTransaction` and wires dbmate to call it before each migration/rollback section; the Postgres driver now issues `SET ROLE` for migrations as well as create/drop, migration table operations, schema dump, and ping. To avoid unsafe behavior, `migrate`/`rollback` now fail with `ErrSetRoleWithNoTransaction` if any involved migration section is marked `transaction:false` when a role is set, and tests were added to cover these cases and validate table ownership under Postgres.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d692724e0fd95d9805f2dd7f941dcaa11da49657. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pkg/dbmate/db_test.go
+++ b/pkg/dbmate/db_test.go
@@ -953,4 +953,3 @@ func TestMigrationsWithRole(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, role, tableOwner)
 }
-


dossy:
@Ragge-dev While I definitely think having some form of pre-apply and post-apply SQL would be a useful mechanism, specifically for your situation/use case, have you considered automatically changing role on login from the Postgres side?

```sql
ALTER ROLE db_migration SET role TO admin;
```

When a user connects with role `db_migration`, Postgres will set their role to `admin`, which effectively should give you the behavior you're looking for without needing an explicit `SET ROLE` during the session.  To demonstrate:

```shell
# psql postgres://postgres:postgres@postgres/dbmate_test?sslmode=disable
psql (17.8 (Debian 17.8-0+deb13u1), server 10.21 (Debian 10.21-1.pgdg90+1))
Type "help" for help.

dbmate_test=# create role db_migration login with password 'migrate';
CREATE ROLE
dbmate_test=# create role admin;
CREATE ROLE
dbmate_test=# grant admin to db_migration;
GRANT ROLE
dbmate_test=# alter role db_migration set role admin;
ALTER ROLE
dbmate_test=# \q

# psql postgres://db_migration:migrate@postgres/dbmate_test?sslmode=disable
psql (17.8 (Debian 17.8-0+deb13u1), server 10.21 (Debian 10.21-1.pgdg90+1))
Type "help" for help.

dbmate_test=> select current_user;
 current_user 
--------------
 admin
(1 row)
```

Ragge-dev:
@dossy Thank you for the tip, I'll investigate your suggestion and get back to you if that solves or use case :star2:

Ragge-dev:
Hello again @dossy!

I have gotten time to look at this again now and the "alter role ..." command seems to solve our use case :heart: Thanks a lot for responding with your suggestion, you have helped us a lot :)

PR #694, opened 2025-10-23
feat: allow passing extra arguments to mysqldump/pgdump

zapotinschii:
feat: allow passing extra arguments to mysqldump/pgdump & fix broken tests (mysql & postgres) 

Allow passing extra arguments to the dump command as follows:
   `dbmate dump -- --flag1 --flag2`
   Everything after `--` will be passed to pgdump or mysqldump.
We can further allow passing extra arguments for other subcommands.

Fixes: https://github.com/amacneil/dbmate/issues/678
Fixes: https://github.com/amacneil/dbmate/issues/688

@amacneil

---

> [!NOTE]
> **Medium Risk**
> Changes the public `Driver` interface and how `dump` shells out to `mysqldump`/`pg_dump`, which could affect downstream driver implementations and schema dump behavior across databases.
> 
> **Overview**
> `dbmate dump` now supports pass-through flags (`dbmate dump -- --flag1 ...`) that are forwarded to `mysqldump`/`pg_dump`, wired from the CLI into a new `DB.Args` field and through the `Driver.DumpSchema(*sql.DB, ...string)` interface.
> 
> MySQL and Postgres drivers were updated to append these extra args to the underlying dump command invocation (other drivers ignore them), README documentation was expanded accordingly, and new MySQL/Postgres tests validate that invalid extra args surface expected tool errors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f5e2e52097f175373a6ad34cd80cba4be6429073. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

Issue #126, opened 2020-03-25
Transaction: false returns error


victoraugustolls:
Hi!

So I'm guessing that using the option `transaction: false` only applies to the first command.
When I have multiple `CREATE INDEX CONCURRENTLY`, it seems like the option only applies to the first operation, and the second fails. Is this expected?

Thanks in advance!

amacneil:
It should apply to the entire migration. Can you share more about the error?

victoraugustolls:
Sure can!

I’m using with Postgres 11.

My first try was creating multiple indexes concurrently without transaction false, which was a mistake and obviously gave me an error.


Second try was setting transaction to false and the error was the same, saying concurrently indexes can’t be created inside a transaction.

Then I tried with only one index creation and it worked as expected.

When I tried again with two or more, the same error occurred, saying concurrently indexes can’t be created inside an transaction.

vrajanap:
I encountered this issue as well - with Postgres 11 on a migration to create multiple indexes. 

After a bit of reading the code, found that dbmate correctly detects that this set of statement needs to be outside transaction and uses [execMigration](https://github.com/amacneil/dbmate/blob/master/pkg/dbmate/db.go#L324) and then later uses the [Exec](https://github.com/amacneil/dbmate/blob/master/pkg/dbmate/db.go#L308) routine. 

This results in `Error: pq: CREATE INDEX CONCURRENTLY cannot run inside a transaction block`.

However, when I tried to repro directly on with lib/pq, it looks like a known limitation - https://github.com/lib/pq/issues/820
```sh
$ psql -h localhost -U postgres -c "create database db_test"
CREATE DATABASE
$ psql -h localhost -U postgres -d db_test -c "create table users(id serial, name text, address text, phonenum bigint)"
CREATE TABLE
$ cat concurrent_index_creation.go
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: concurrent_index_creation.go
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ package main
   2   │
   3   │ import (
   4   │     "database/sql"
   5   │     "fmt"
   6   │
   7   │     _ "github.com/lib/pq"
   8   │ )
   9   │
  10   │ func main() {
[...]

PR #696, opened 2025-10-25
Migrate from lib/pq to pgx

kishaningithub:


kishaningithub:
@amacneil The test for db dump from post gres is failing because `\\unrestrict mqJlN2GMO0T4PWYvE3MQmVGGOeL8bYZzubTrpPdsVTqjfWLIC9YHl3RMpk0g1bH` is getting added in the pg_dump output (not sure why)

## Exact DIFF (Screenshot below)

image

## From pipeline (Screenshot below)

https://github.com/amacneil/dbmate/actions/runs/18802140315/job/53651080308?pr=696

image

kishaningithub:
@amacneil Also eventhough i updated only the postgres part of this codebase some mysql parts are also failing, can you help in looking into that as well?

kishaningithub:
Related - https://github.com/amacneil/dbmate/pull/677

zapotinschii:
Both postgres and mysql tests are fixed here btw: https://github.com/amacneil/dbmate/pull/694

dossy:
@kishaningithub (and anyone else who would like to see this change happen), can you articulate any benefits there would be in making this change?

The argument to use `pgx` over `lib/pq` because `lib/pq` is in maintenance mode isn't very compelling: `lib/pq` is still actively maintained and is not abandonware, as there was even [a commit a few hours ago today, 2026-01-11](https://github.com/lib/pq/commit/99baf2fb55f2b6cc2faa67ad6b291bbf6393200f).

I see `lib/pq` as very safe and stable, as it hasn't had a single CVE in the 13+ years it's been around ([first commit in GitHub in 2012](https://github.com/lib/pq/commit/df1f1317967381f224717825633bb31354d8767c)), is pure Go (no external dependencies), and being in maintenance mode means it should be very stable and predictable at this point.

`pgx`, on the other hand, has had [two CVEs from 2024](https://github.com/jackc/pgx/security).  Granted, the two CVEs are SQL injections and the primary use case for `dbmate` generally doesn't involve unsanitized user-supplied inputs, but it speaks to the complexity of `pgx`, that such issues existed.

`pgx` sure is fancy and has a lot of bells and whistles, but I don't see any of them being necessary for what `dbmate` requires, but that might just be my lack of knowledge of what `dbmate` currently cannot do because it's using `lib/pq` and not `pgx`.

Is there a use case for `dbmate` that is currently not possible because it uses `lib/pq`?

kishaningithub:
pgx works beautifully well with cockroach db but lib/pq does not

dossy:
Related: #595

dossy:
> pgx works beautifully well with cockroach db but lib/pq does not

Can you demonstrate what you mean with a reproducible example?

Because the CockroachDB documentation seems to imply that [the Go `pq` driver is fully supported](https://www.cockroachlabs.com/docs/stable/install-client-drivers?filters=go#pq).

steelbrain:
The pq driver has this very annoying limitation: https://github.com/amacneil/dbmate/issues/126#issuecomment-672284761

Would it be elevated by moving to pgx? Got bit by that bug today

dossy:
@steelbrain That's a great question.  You could try experimenting using the change in this PR and determine whether it would or not.

PR #699, opened 2025-11-01
Bump go-sqlite3 to 1.14.32

brettatoms:
This fixes using fts5 in sqlite migrations on Linux.  

The fts5 tests were passing on Linux before this PR but when I tried to run a migration that used fts5 I would get the `Error: no such module: fts5` error.  With this change it works but TBH I'm not 100% sure what change in go-sqlite3 fixed it.


brettatoms:
Had the same issue on MacOS and this PR also fixes it there.

brettatoms:
I'm going to close this.  I think bundling the sqlite library into dbmate is the wrong approach since you won't be able to dynamically load sqlite extensions that were bundled with dbmate.

PR #701, opened 2025-11-13
Do not create migrations table on dump

MatthiasKunnen:
When running `dbmate dump`, no changes to the database should be made.

Use case: running `dbmate dump` with credentials that do not have _CREATE_ permissions.

### DIFF
--- a/pkg/driver/mysql/mysql.go
+++ b/pkg/driver/mysql/mysql.go
@@ -158,6 +158,14 @@ func (drv *Driver) mysqldumpArgs() []string {
 }
 
 func (drv *Driver) schemaMigrationsDump(db *sql.DB) ([]byte, error) {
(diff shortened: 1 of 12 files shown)


pkg/driver/mysql/mysql.go
```diff
@@ -158,6 +158,14 @@ func (drv *Driver) mysqldumpArgs() []string {
 }
 
 func (drv *Driver) schemaMigrationsDump(db *sql.DB) ([]byte, error) {
+	exists, err := drv.MigrationsTableExists(db)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if migration table exists: %w", err)
```
dossy: See previous comment about returning `err` as-is.

PR #703, opened 2025-11-15
Support requiring SSL, and verifying CA, for MySQL

MatthiasKunnen:
This PR adds support for requiring a secure connection and man-in-the-middle protection in the form of the `DBMATE_MYSQL_SSL_MODE` and `DBMATE_MYSQL_CA_PATH` parameter.

- `DBMATE_MYSQL_SSL_MODE` is an implementation of <https://dev.mysql.com/doc/refman/8.4/en/connection-options.html#option_general_ssl-mode>
- `DBMATE_MYSQL_CA_PATH` implements <https://dev.mysql.com/doc/refman/8.4/en/connection-options.html#option_general_ssl-ca>

This PR will also fix tests failing on the main branch due to certificate verification problems.

Important caveat, the CLI parameter `--ssl-mode` used for mysqldump is not present in mariadb dump. If there is a way to detect whether mariadb-dump is used, I could change the parameters accordingly but there does not seem to be a distinction at the moment. Suggestions welcome.

I'll refrain from further work until there is some feedback on the approach and whether there is interest in this change.

Todo:
- [ ] Decide how to handle mariadb dump compatibility
- [ ] Add documentation

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,7 @@ services:
       CLICKHOUSE_CLUSTER_01_TEST_URL: clickhouse://ch-cluster-01:9000/dbmate_test
       CLICKHOUSE_CLUSTER_02_TEST_URL: clickhouse://ch-cluster-02:9000/dbmate_test
       MYSQL_TEST_URL: mysql://root:root@mysql/dbmate_test
+      DBMATE_MYSQL_SSL_MODE: DISABLED
       POSTGRES_TEST_URL: postgres://postgres:postgres@postgres/dbmate_test?sslmode=disable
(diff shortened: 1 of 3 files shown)


dossy:
The change in #730 addresses very basic SSL on/off control for dbmate with MySQL and MariaDB.  No support for CA certificates was implemented there.

Personally, my opinion is rather than trying to shoehorn a menagerie of configuration knobs onto `dbmate`, it would be far more useful to add a general-purpose "extra args" value that can be specified that is passed through to the underlying dump executable, then the `dbmate` user has full control now and in the future without additional changes to `dbmate` itself.

dossy:
Also: it would be nice to better support CA certificate handling for SSL/TLS connections for all our drivers, not just MySQL.  What I mean is, I would prefer a solution that was generalized at the interface level, and each driver implements the necessary mechanism to provide the functionality.

PR #704, opened 2025-11-15
Update tests to deal with \unrestrict in postgres 17's dump output

MatthiasKunnen:
See https://www.postgresql.org/docs/17/release-17-6.html and https://www.postgresql.org/support/security/CVE-2025-8714/.


sofuture:
Thanks for this @MatthiasKunnen -- I ended up PRing it separately because of github permissions being annoying, but this fix is in as of #716 .

PR #785, opened 2026-04-12
fix: redact full password when it contains @ sign

### LINKED ISSUE #784: DBmate prints password if unable to parse string
**Description**
I have a conn

- Version: 2.32.0
- Database: postgreSQL
- Operating System: Alpine

**Steps To Reproduce**

Setup connection string as env var like so `postgres://username:oRAND44@W)£+$\$r{5F?#x:$A\y^9\;&pzIS1&b_k@1.1.1.1:5432/db_name`
Try to run dbmate 

**Expected Behavior**
If parsing fails parts of the password are not printed in the logs

**Actual Behavior**
The following is logged

> Error: parse "postgres://username:********@W)£+$\\$r{5F?": invalid character "\\" in host name

**Note**

While I understand that apparently I need to do some URL encoding of my password to make it work, I still expect there to be an option to NOT log passwords/connection string whatsoever.

Yanhu007:
Fixes #784

## Problem

`redactLogString` uses `[^@]+` to match the password portion, stopping at the first `@`. When a password contains `@` (e.g. `pass@word`), only the part before the first `@` is redacted, leaking the rest in error messages:

```
Input:  postgres://user:oRAND44@leaked_part@host:5432/db
Output: postgres://user:********@leaked_part@host:5432/db  ← password leaked!
```

## Fix

Change `[^@]+` to `.+` (greedy) so the regex matches everything up to the **last** `@`, redacting the entire password:

```
Output: postgres://user:********@host:5432/db  ← fully redacted
```

Added test case for passwords containing `@`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -343,7 +343,7 @@ func getDatabaseURL(c *cli.Context) (u *url.URL, err error) {
 
 // redactLogString attempts to redact passwords from errors
 func redactLogString(in string) string {
-	re := regexp.MustCompile("([a-zA-Z]+://[^:]+:)[^@]+@")
+	re := regexp.MustCompile("([a-zA-Z]+://[^:]+:).+@")
 
 	return re.ReplaceAllString(in, "${1}********@")
 }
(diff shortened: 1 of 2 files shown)


sofuture:
Will this overmatch if there are `@` later in the connection string for some reason? i.e. `postgresql://user:password@host:5432/mydb?foo=b@r`

dossy (approved):

PR #790, opened 2026-04-29
v2.33.0

sofuture:
Release v2.33.0


dossy:
@amacneil, @sofuture --

Just a thought: given the increase in supply-chain attacks, how do you all feel about _not_ making dependency updates a standard part of the release process, and instead, we update dependencies manually as needed with validation that the version being updated to has been around long enough to rule out some percent chance of being a supply chain attack?

sofuture:
That's almost certainly the right approach to take. I will update release guidelines.

PR #792, opened 2026-04-29
Simplify release process by removing dependency upgrade steps

sofuture:
Removed instructions for upgrading dependencies from the release process.


dossy (approved):
re: https://github.com/amacneil/dbmate/pull/790#issuecomment-4344304153

PR #795, opened 2026-05-13
fix: report dedicated error for multiple migrate:down annotations

SAY-5:
When a migration section contains more than one `-- migrate:down` annotation, the parser returns `ErrParseMissingUp` ("missing up block"), which is misleading because the up block is present. This adds `ErrParseMultipleDown` so the reported failure matches what the user typed.

### DIFF
--- a/pkg/dbmate/migration.go
+++ b/pkg/dbmate/migration.go
@@ -74,6 +74,7 @@ var (
 	ErrParseMissingDown    = errors.New("dbmate requires each migration to define a down block with '-- migrate:down'")
 	ErrParseWrongOrder     = errors.New("dbmate requires '-- migrate:up' to appear before '-- migrate:down'")
 	ErrParseUnexpectedStmt = errors.New("dbmate does not support statements preceding the '-- migrate:up' block")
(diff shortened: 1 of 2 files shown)


pkg/dbmate/migration.go
```diff
@@ -74,6 +74,7 @@ var (
 	ErrParseMissingDown    = errors.New("dbmate requires each migration to define a down block with '-- migrate:down'")
 	ErrParseWrongOrder     = errors.New("dbmate requires '-- migrate:up' to appear before '-- migrate:down'")
 	ErrParseUnexpectedStmt = errors.New("dbmate does not support statements preceding the '-- migrate:up' block")
+	ErrParseMultipleDown   = errors.New("dbmate does not support multiple '-- migrate:down' annotations in a single migration section")
```
dossy: I think the spirit of this change has its heart in the right place, but this message may be misinterpreted, and introduces language ("migration section") which isn't obvious.

The error message should be along the lines of "every `-- migrate:down` must be preceeded by a `-- migrate:up`" which is succinct and tells the user exactly why the migration file is being rejected and how to correct the issue.

SAY-5:
Reworded to `dbmate requires every '-- migrate:down' to be preceded by a '-- migrate:up'`, matching the existing error style and stating the fix directly.

PR #814, opened 2026-07-10
Replace broken Homebrew bump action with brew bump-formula-pr

hillna:
## Summary

- Replace `mislav/bump-homebrew-formula-action` with Homebrew's own `brew bump-formula-pr` CLI, which correctly follows GitHub's HTTP 303 tarball redirects ([mislav/bump-homebrew-formula-action#340](https://github.com/mislav/bump-homebrew-formula-action/issues/340))
- Use `Homebrew/actions/setup-homebrew` with `core: true` so the `dbmate` formula is available on Linux CI runners
- Add `workflow_dispatch` input to manually retry a Homebrew bump without cutting a new release tag
- Preserve the existing `dbmate-releasebot` fork-and-PR flow via `--fork-org=dbmate-releasebot` and `RELEASE_TOKEN`

## Test plan

- [x] Triggered `workflow_dispatch` on this branch with `tag=v2.34.1` ([run #29102169692](https://github.com/amacneil/dbmate/actions/runs/29102169692)) — no HTTP 303 error; failed as expected because v2.34.1 is already in homebrew-core via BrewTestBot autobump
- [ ] After merge, confirm the next release tag opens a PR to `Homebrew/homebrew-core` from `dbmate-releasebot`

Made with [Cursor](https://cursor.com)

PR #817, opened 2026-07-20
fix: remove extra bound argument in BigQuery InsertMigration

igormakhtes:
`InsertMigration` in the BigQuery driver interpolates `version` into the SQL string, then also passes it to `db.Exec` as a bound argument. The query has no placeholders, so the argument has nothing to bind to. Real BigQuery silently ignores it, but newer versions of goccy/bigquery-emulator reject the insert:

```
googleapi: Error 400: failed to exec INSERT INTO ... : sql: expected 0 arguments, got 1
```

This makes `dbmate up` fail against the emulator right after applying the first migration. `DeleteMigration` does the same interpolation without the extra argument, so this looks like a leftover from #523.

Following review, both `InsertMigration` and `DeleteMigration` now bind `version` as a query parameter instead of interpolating it into the SQL, and `InsertMigration` runs on the transaction `Migrate()` passes in (as `DeleteMigration` already did).

Tested the `pkg/driver/bigquery` suite against emulator 0.4.4 (the version pinned in docker-compose.yml, which tolerates the extra arg) and 0.8.1 (which rejects it) - passes on both. CI doesn't catch this today because it pins 0.4.4; happy to bump the image here or in a follow-up if you want the regression covered.

### DIFF
--- a/pkg/driver/bigquery/bigquery.go
+++ b/pkg/driver/bigquery/bigquery.go
@@ -318,7 +318,7 @@ func (drv *Driver) InsertMigration(_ dbutil.Transaction, version string) error {
 
 	queryTemplate := `INSERT INTO %s.%s (version) VALUES ('%s');`
 	queryString := fmt.Sprintf(queryTemplate, config.dataSet, drv.migrationsTableName, version)
-	_, err = db.Exec(queryString, version)
+	_, err = db.Exec(queryString)
 	if err != nil {
 		return err
 	}
--- a/pkg/driver/bigquery/bigquery.go
+++ b/pkg/driver/bigquery/bigquery.go
@@ -283,7 +283,7 @@ func (drv *Driver) MigrationsTableExists(db *sql.DB) (bool, error) {
 	return exists, nil
 }
 
-func (drv *Driver) DeleteMigration(util dbutil.Transaction, version string) error {
+func (drv *Driver) DeleteMigration(tx dbutil.Transaction, version string) error {
 	db, err := drv.Open()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pkg/driver/bigquery/bigquery.go
```diff
@@ -318,7 +318,7 @@ func (drv *Driver) InsertMigration(_ dbutil.Transaction, version string) error {
 
 	queryTemplate := `INSERT INTO %s.%s (version) VALUES ('%s');`
 	queryString := fmt.Sprintf(queryTemplate, config.dataSet, drv.migrationsTableName, version)
-	_, err = db.Exec(queryString, version)
```
dossy, line 321: It would be better if we actually did use the bind parameter instead of relying on string interpolation which could be a SQL injection vector given the correct wrong circumstances.

We should similarly fix the `DeleteMigration` query to properly use a bind parameter, too.

Would you please make this change?

dossy (changes requested):

dossy:
It would also be good to fix `InsertMigration()` so that it executes the query using the passed-in `dbutil.Transaction` that [`Migrate()` creates for executing migration queries](https://github.com/igormakhtes/dbmate/blob/bb656c43b5abd1ff0dc1c67ca4e885cfc6cee098/pkg/dbmate/db.go#L387-L415).  `DeleteMigration()` already does this.

igormakhtes:
Thanks for the review @dossy ! 

I’ve pushed the requested changes:
* Updated `InsertMigration` and `DeleteMigration` to use bind parameters
* Updated `InsertMigration()` to execute through the transaction passed into Migrate()

dossy (approved):
Thanks, this looks good.

AlonMenczerLema:
Thanks @dossy !

When can we expect a release?

dossy:
I'm not sure, it looks like #814 was trying to release v2.34.2.  Once that's sorted out, it looks like that'll trigger the next release.

PR #837, opened 2026-09-11
Document environment variable usage in README (#500)

### LINKED ISSUE #500: Add a section to the README documenting how environment variables are used by dbmate
As a consequence of #491, it would be helpful to add a section to the README that describes and explains how dbmate uses environment variables.

* Which settings are derived from which environment variables
* How dbmate loads from `.env` in the current working directory by default
* How to specify one or more dot-env files using the `--env-file` option
* Variable precedence: dot-env files, the calling process's environment, command line option value

MrBeldum:
## Summary
Adds a README section documenting how dbmate uses environment variables, as requested in #500 (follow-up to #491 / `--env-file`).

The new **Environment Variables** section covers:
- Which settings map to which environment variables (`DATABASE_URL`, `DBMATE_*`, plus `DBMATE_VERBOSE`)
- Default loading of `.env` from the current working directory
- `--env-file` for one or more dotenv files (and that specifying it replaces the default `.env` unless included)
- Precedence: CLI flags > process environment > dotenv files > built-in defaults
- Database URL specifics: `--url` over `--env` / `DATABASE_URL`

Also adds a short cross-link from **Connecting to the Database**, and a TOC entry.

Docs only; behavior verified against `main.go` (`loadEnvFiles`, flag `EnvVars`, `getDatabaseURL`) and `main_test.go`.

Closes #500.

## Test plan
- [ ] Skim the new section for accuracy against `main.go` / `main_test.go`
- [ ] Confirm TOC link and Connecting cross-link resolve
- [ ] Confirm examples match existing README style


dossy (approved):
Nice and concise.  Would like to see if anyone else still has questions that aren't answered by this that should be included, but I think this is fine to merge as-is.

